PR #15, opened 2024-12-23
add models_download.py script to download model files at docker-building time, to avoid model download issues at run time.

mazhengm2:


dolfim-ibm:
This was done also in https://github.com/DS4SD/docling-serve/pull/24 and https://github.com/DS4SD/docling-serve/pull/17

PR #16, opened 2025-01-02
Add examples for options and posting a local file for conversion

nerdalert:


dolfim-ibm:
The example is now in #17 .

PR #17, opened 2025-01-09
api v1alpha1

guimou:
Rework for API version v1alpha1 as discussed.


PeterStaar-IBM:
@dolfim-ibm Let's prioritize this PR asap!

dolfim-ibm:
@guimou I started with a few changes
- used the Annotated syntax for the parameters
- use default values and not `Optional` (which in Python means "can be `None`" and has almost no relation with the concept of optional)
- refactor the request arguments logic. where possible use the native enum.
- make the gradio part optional (but highly advised)
- rebase with main such that tests build the image


Next, we will look
- the background processing
- input/output schemas

guimou:
Just had a look at the changes, that's great. Thanks a lot!

dolfim-ibm:
@guimou I reworked a few parts, I think that now the PR is ready for merge and release.

@vishnoianil do you want to also have a quick look?

dolfim-ibm (dismissed):
great

Containerfile
```diff
@@ -1,32 +1,61 @@
...
+        poetry install --no-root --no-cache --no-interaction --all-extras --without dev; \
+    fi && \
+    echo "Downloading models..." && \
+    python models_download.py && \
+    chown -R 1001:0 /opt/app-root/src && \
+    chmod -R g=u /opt/app-root/src
+
+COPY --chown=1001:0 --chmod=664 ./docling_serve ./docling_serve
 
-EXPOSE 5000
```
vishnoianil, line 30: is there any specific need to run this service on port 8080? I am wondering If we can revert it 5000 or some other non-80** port? @dolfim-ibm

vishnoianil:
Awesome work @guimou 

@dolfim-ibm minor comment, PR looks good to me. I am wondering can we enable the arm64 arc for "cpu only" image ? It's enabled for GPU image, but not for CPU only image. I remember we were discussing about some dependencies not available for arm64, and that was causing issue in building arm64 image, is that still true?

dolfim-ibm:
I realized we missed addressing the input http with custom headers. I'm making a pushing a version similar to what we have on origin/main.

dolfim-ibm:
@vishnoianil The PORT has been changed to 5001. Additionally I also reworked the input payload to match the format already in `main` which allows custom headers and base64 input.

PeterStaar-IBM (dismissed):
let's go!

dolfim-ibm (approved):

PeterStaar-IBM (approved):
let's go! 🚀

PR #29, opened 2025-02-07
docs: correct multipart/form-data curl example

archasek:
Minor fix, as curl didn't work with arrays or concatenated strings sent.


archasek:
Will create a new cleaner one

PR #31, opened 2025-02-07
fix(gradio-ui): correct output row width, prevent options row from hiding

archasek:
1. Options row was collapsing on each processing. Now they are collapsed only on startup.

2. Output row width was incorrect. Changed that to tabs.

Old:
image

New:
image

### DIFF
--- a/docling_serve/gradio_ui.py
+++ b/docling_serve/gradio_ui.py
@@ -396,7 +396,7 @@ with gr.Blocks(
                 file_reset_btn = gr.Button("Reset", scale=1)
 
     # Options
-    with gr.Accordion("Options") as options:
+    with gr.Accordion("Options", open=False) as options:
         with gr.Row():
             with gr.Column(scale=1):
                 to_formats = gr.CheckboxGroup(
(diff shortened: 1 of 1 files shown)


guimou:
- Thanks for the UI glitch in the output. It did not happen locally, so don't what was going on...
- Options panel was opened on startup on purpose to show which options were available. @dolfim-ibm wanted Embedded images to be the default in the API, but I think the UI will be used mostly by people wanting to quickly check the Markdown, which can be made unreadable with the base64 code in the middle. So you can directly check/modify the options at startup, then the options would collapse when you launch the first processing.
But that's not a hill I would die on, either behaviour is fine by me.

dolfim-ibm:
I also think the options should be opened in the beginning, and I'm still for having the embedded behavior as default.

What I think we should consider is to have some ENV which allows to change the defaults.

archasek:
I agree with:
1. Options panel should be always open by default and not be collapsing on each processing. It's annoying to have to open it every time when we open the UI.
2. Placeholder option for images turned on by default. Same as above - I need to always switch it on manually.
3. I'm in favor of having a single source of truth for all settings for the backend app and ui, perhaps in .env file(s). At the moment some of that is distributed - start_server.sh, docling_serve/.env, Containerfile. Would be great to work same way for both local and docker builds.

Happy to finish 1 & 2 in this PR if you agree?

archasek:
I think it's no longer needed?

PR #35, opened 2025-02-08
Fix __main__ when workers > 1 or reload enabled

guimou:
When reload is set, or when uvicorn workers are more than one, the application cannot start.
That's because Uvicorn requires the application to be passed as an import string (e.g., 'docling_serve.app:app'). This is because Uvicorn needs to manage multiple processes or watch for file changes, which requires knowing how to import the application.

### DIFF
--- a/docling_serve/__main__.py
+++ b/docling_serve/__main__.py
@@ -1,6 +1,5 @@
 import os
 
-from docling_serve.app import app
 from docling_serve.helper_functions import _str_to_bool
 
 # Launch the FastAPI server
@@ -10,8 +9,9 @@ if __name__ == "__main__":
     port = int(os.getenv("PORT", "5001"))
     workers = int(os.getenv("UVICORN_WORKERS", "1"))
     reload = _str_to_bool(os.getenv("RELOAD", "False"))
+
     run(
-        app,
+        "docling_serve.app:app",
         host="0.0.0.0",
         port=port,
         workers=workers,


docling_serve/__main__.py
```diff
@@ -10,11 +10,20 @@
...
-    run(
-        app,
-        host="0.0.0.0",
-        port=port,
-        workers=workers,
-        timeout_keep_alive=600,
-        reload=reload,
-    )
+
+    if workers > 1 or reload:
```
archasek, line 14: is it needed? won't this work even if workers = 1?

guimou:
Did not test with one, but this does not work with workers = 3 on
Openshift, this is how I discovered it.
I need 3 to maximize GPU usage. Otherwise the Pod monopolizes the whole GPU
(a T4 in my case), only using a small part of it.
More Pods would not work as they would each need their own GPU, again only
using part of it.

docling_serve/__main__.py
guimou: Yes, I just tested it, always calling the module whatsoever works.
I put it like this originally not to interfere too much with the rest, but can totally modify for a more generic approach.
Btw, the start_server.sh is already written like this.
Pushing the change.

docling_serve/__main__.py
archasek, line 1: this line can be safely removed :)

archasek:
1. `from docling_serve.app import app` this line can be safely removed :)
2. sign off commits please https://github.com/DS4SD/docling-serve/pull/35/checks?check_run_id=36917148709

guimou:
1. Removed line
2. All commits were signed off. I don't know what's going on... I rebased to have only one commit, also signed-off.

archasek:
> 2. All commits were signed off. I don't know what's going on... I rebased to have only one commit, also signed-off.

Yeah I can see that. Have a look at (at the bottom) https://github.com/DS4SD/docling-serve/runs/36923497614 
Btw. you may want to read about keeping emails private too: https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-email-preferences/setting-your-commit-email-address#setting-your-commit-email-address-on-github

guimou:
I must be dumb. My email is already private. I tried to use the anonymous versions "3944034+guimou@users.noreply.github.com" or "guimou@users.noreply.github.com" as explained in the doc, but no luck...

dolfim-ibm:
Shouldn't we simply recommend the direct usage of uvicorn instead of maintaining some custom logic for it?

For example using this script for launching: https://github.com/DS4SD/docling-serve/blob/main/start_server.sh

guimou:
> Shouldn't we simply recommend the direct usage of uvicorn instead of maintaining some custom logic for it?
> 
> For example using this script for launching: https://github.com/DS4SD/docling-serve/blob/main/start_server.sh

Yes, we can totally run uvicorn directly. You will just have to pass all the launch parameters.
If we use a script to pass the parameters, it cannot directly be `start_server.sh` as it uses poetry to launch uvicorn. However in the container we already are in the virtual environment (under /opt/app-root), so there is no need for it to switch context. Uvicorn can be called directly.

dolfim-ibm (approved):

PR #38, opened 2025-02-10
fix: Set root UI path when behind proxy

guimou:
There was one parameter missing in the Gradio  config to make it work properly behind a proxy, like the HAProxy router in an OpenShift Route.


dolfim-ibm (approved):
lgtm

PR #41, opened 2025-02-12
ci: add semantic release and build/publish python wheel

dolfim-ibm:
- Add standard GH templates (PR and SECURITY)
- Add mergify with checks for conventional commits in PR titles
- Reuse jobs for building images
- Add publish to pypi on releases
- Add semantic release automation

Releases are triggered manually from Actions > Run CD.


.github/workflows/images.yml
```diff
@@ -7,190 +7,52 @@ on:
...
-        id: quay_serve_gpu_meta
-        uses: docker/metadata-action@v5
-        with:
-          images: ${{ env.QUAY_REGISTRY }}/${{ env.QUAY_DOCLING_SERVE_GPU_IMAGE_NAME }}
+    uses: ./.github/workflows/job-image.yml
+    with:
+      publish: true
+      environment: registry-creds
+      build_args: |
+        --build-arg CPU_ONLY=true
```
vishnoianil, line 55: i think CPU_ONLY should be false here.

.github/workflows/job-image.yml
```diff
@@ -0,0 +1,134 @@
...
+          restore-keys: |
+            ${{ runner.os }}-buildx-
+
+      - name: Extract metadata (tags, labels) for docling-serve ghcr image
+        id: ghcr_meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ${{ env.GHCR_REGISTRY }}/${{ inputs.ghcr_image_name }}
+
+      - name: Build and push docling-serve-cpu image to ghcr.io
```
vishnoianil, line 84: I think we can get rid of "docling-serve-cpu image" part in the name, because the same job will run for both docling-serve-cpu and docling-serve. Or we can probably get rid of this entirely, because the caller job has the name anyways?

vishnoianil: or just says "Build and push image to ghcr.io"?

vishnoianil (commented):
@dolfim-ibm minor comments. I think we can get rid of the commented out code and this PR is good to go.

vishnoianil (dismissed):

vishnoianil (approved):

PR #46, opened 2025-02-18
feat(config): add global document_timeout setting

archasek:
1. Add support for document_timeout docling setting. Can be set globally (env var) or in the request payload.
2. Move .env from docling_serve folder to root.
3. Fix error handling, as there was serialization problem with the old code.
API - When there's timeout, this kind of response (500) is returned:
```
{
    "detail": {
        "errors": [
            {
                "component_type": "document_backend",
                "module_name": "NoneType",
                "error_message": "Page 4 failed to parse."
            },
            {
                "component_type": "document_backend",
                "module_name": "NoneType",
                "error_message": "Page 5 failed to parse."
            },
            ...
```

Gradio:
image

Logs:
```
WARNING:        2025-02-19 01:23:06,856 - docling.pipeline.base_pipeline - Document processing time (8.141 seconds) exceeded the specified timeout of 3.000 seconds
INFO:     127.0.0.1:53490 - "POST /v1alpha/convert/file HTTP/1.1" 500 Internal Server Error
ERROR:  2025-02-19 01:23:07,164 - docling_serve.gradio_ui - Error processing file: No documents were generated by Docling.
```


docling_serve/settings.py
```diff
@@ -1,6 +1,9 @@
 from pydantic_settings import BaseSettings, SettingsConfigDict
-
+from typing import Optional
 
 class Settings(BaseSettings):
-
-    model_config = SettingsConfigDict(env_prefix="DOCLING_")
+    model_config = SettingsConfigDict(
+        env_prefix="DOCLING_",
+        env_file="docling_serve/.env",
```
dolfim-ibm, line 7: shouldn't this be simple `.env`? otherwise we force the user on the given location.

archasek: I initially thought there should be a single `.env` file in the root folder for the entire project. However, I noticed that there is already a `.env.example` file in the `docling_serve` folder. So, I decided not to change the current design at this point. What do you think?

dolfim-ibm: I think it is better with a single `.env`. Let's align all options, and in case use prefixes.

archasek: Done, please have a look dolfim. I also noticed an issue with error handling and fixed that.

docling_serve/response_preparation.py
```diff
@@ -69,9 +72,13 @@ def _export_document_as_content(
         if export_doctags:
             document.doctags_content = new_doc.export_to_document_tokens()
     elif conv_res.status == ConversionStatus.SKIPPED:
-        raise HTTPException(status_code=400, detail=conv_res.errors)
+        raise HTTPException(
```
dolfim-ibm, line 75: On the single document, this makes perfect sense.
But this might not be the expected behavior when the user provides multiple document.

In this proposal, if 10 documents are provided and 1 fails, it will raise the error and produce no results.

In the latter, it would be better to have some manifest response file in the zip, where the system lists the result for each file.

dolfim-ibm:
Superseded by #142

archasek:
@dolfim-ibm are you planning to implement such timeout param? That's probably the only reason I had to fork the repo.

dolfim-ibm:
it should be done already. did you look at the PR above?

archasek:
yes yes, thanks a lot

PR #48, opened 2025-02-18
fix: support python 3.13 and docling updates and switch to uv

dolfim-ibm:
- Update to new version of Docling
- Use the `docling-tools models download` utility
- Enable support for Python 3.13
- Use `uv` to manage packages

Note: the gpu image is built only for linux/amd64, because the torch version doesn't exist for arm64.


cau-git (approved):
🚀 🎸

PR #49, opened 2025-02-19
chore: use uv in start_server.sh and update docs

Fogapod:
`start_server.sh` and docs were not updated after switching to uv.

I also added a paragraph on extra explaining how uv extras work.

`CONTRIBUTING.md` still mentions poetry. I haven't touched it because I'm not sure what the workflow is and what parts should be kept


dolfim-ibm (approved):
lgtm, thanks!

dolfim-ibm:
Note that we are almost done with #47 which will obsolete the `start_server.sh` script.

dolfim-ibm (approved):

dolfim-ibm (approved):

PR #58, opened 2025-02-21
refactor: Use bytes as options key

Fogapod:
`hexdigest` is a little bit slower and there is no reason to use strings here.

even if you needed to be able to copypaste these keys for some reason you still can: `{b'\xf6\x9bg\xef\xd8\xa8\x8be\x92&\xbc\xaeq\x10\xaf^L\xbd\xd4\xaf': <docling.document_converter.DocumentConverter object at 0x7598596dacc0>}`


dolfim-ibm (approved):
lgtm

PR #60, opened 2025-02-21
feat: Async api

### LINKED ISSUE #51: First version of the async api
Features to provide:
- Submit a task and return a `task_id`
- Check the status of the task
- When it is successful, endpoint to retrieve the result

Async compute:
- Simple queue in the same process
- Prepare design for the other async compute engines

dolfim-ibm:
Introduce the Async compute API

```
# sync (currently in place)
/v1alpha/convert/source
/v1alpha/convert/file

# launch async
/v1alpha/convert/source/async
/v1alpha/convert/file/async

# check status
/v1alpha/status/poll/{task_id}?wait=10
/v1alpha/status/ws/{task_id}

# get result
/v1alpha/result/{task_id}
```

Still TODO

- [x] Definition of async api specs
- [x] Implement first option with a local queue
- [x] Websockets for push updates
- [ ] Deal with file return type (with `return_as_file` or multiple inputs)

Resolves #51

### DIFF
--- a/docling_serve/app.py
+++ b/docling_serve/app.py
@@ -14,6 +14,7 @@ from fastapi import (
     Query,
     UploadFile,
     WebSocket,
+    WebSocketDisconnect,
 )
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import RedirectResponse
@@ -29,7 +30,9 @@ from docling_serve.datamodel.requests import (
 from docling_serve.datamodel.responses import (
     ConvertDocumentResponse,
     HealthCheckResponse,
+    MessageKind,
     TaskStatusResponse,
+    WebsocketMessage,
 )
 from docling_serve.docling_conversion import (
     convert_documents,
@@ -81,7 +84,6 @@ _log = logging.getLogger(__name__)
 # Context manager to initialize and clean up the lifespan of the FastAPI app
 @asynccontextmanager
 async def lifespan(app: FastAPI):
-
     # Converter with default options
     pdf_format_option, options_hash = get_pdf_pipeline_opts(ConvertDocumentsOptions())
     converters[options_hash] = DocumentConverter(
@@ -113,6 +115,25 @@ async def lifespan(app: FastAPI):
     #     gradio_ui.close()
 
 
+class ConnectionManager:
+    def __init__(self):
+        self.active_connections: list[WebSocket] = []
+
+    async def connect(self, websocket: WebSocket):
+        await websocket.accept()
+        self.active_connections.append(websocket)
+
+    def disconnect(self, websocket: WebSocket):
+        self.active_connections.remove(websocket)
+
+    async def send_personal_message(self, message: str, websocket: WebSocket):
+        await websocket.send_text(message)
+
+    async def broadcast(self, message: str):
+        for connection in self.active_connections:
+            await connection.send_text(message)
+
+
 ##################################
 # App creation and configuration #
 ##################################
@@ -138,7 +159,6 @@ def create_app():  # noqa: C901
 
     # Mount the Gradio app
     if docling_serve_settings.enable_ui:
-
         try:
             import gradio as gr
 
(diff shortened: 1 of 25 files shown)


docling_serve/datamodel/engines.py
```diff
@@ -0,0 +1,30 @@
...
+class Task(BaseModel):
+    task_id: str
+    task_status: TaskStatus = TaskStatus.PENDING
+    request: Optional[ConvertDocumentsRequest]
+    result: Optional[ConvertDocumentResponse] = None
+
+    def is_completed(self) -> bool:
+        if self.task_status in [TaskStatus.SUCCESS, TaskStatus.FAILURE]:
+            return True
+        return False
```
Fogapod, line 30:
```suggestion
        return self.task_status in [TaskStatus.SUCCESS, TaskStatus.FAILURE]
```

docling_serve/engines/async_local/worker.py
```diff
@@ -0,0 +1,131 @@
...
+
+    async def loop(self):
+        _log.debug(f"Starting loop for worker {self.worker_id}")
+        while True:
+            task_id: str = await self.orchestrator.task_queue.get()
+            self.orchestrator.queue_list.remove(task_id)
+
+            if task_id not in self.orchestrator.tasks:
+                raise RuntimeError(f"Task {task_id} not found.")
+            task = self.orchestrator.tasks[task_id]
```
Fogapod, line 38: shouldnt this be a part of `AsyncLocalOrchestrator` API? too easy to make a mistake with 3 separate properties

docling_serve/datamodel/engines.py
Fogapod: I think ruff's `SIM` would suggest this fix

docling_serve/engines/async_local/orchestrator.py
```diff
@@ -0,0 +1,101 @@
...
+
+class TaskNotFoundError(OrchestratorError):
+    pass
+
+
+class AsyncLocalOrchestrator(BaseOrchestrator):
+    def __init__(self):
+        self.task_queue = asyncio.Queue()
+        self.tasks: Dict[str, Task] = {}
+        self.queue_list: List[str] = []
```
cau-git, line 34: Maybe one could encapsulate these three members in a class that derives from asyncio.Queue to make it cleaner and less error prone to use.

cau-git (commented):
✨  Beautiful! Only small comments below.

docling_serve/app.py
```diff
@@ -72,19 +95,51 @@ async def lifespan(app: FastAPI):
...
+        self.active_connections: list[WebSocket] = []
+
+    async def connect(self, websocket: WebSocket):
+        await websocket.accept()
+        self.active_connections.append(websocket)
+
+    def disconnect(self, websocket: WebSocket):
+        self.active_connections.remove(websocket)
+
+    async def send_personal_message(self, message: str, websocket: WebSocket):
```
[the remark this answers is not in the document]
dolfim-ibm: good catch, that was a placeholder from the fastapi docs

docling_serve/engines/async_local/orchestrator.py
dolfim-ibm: I think this and other refactoring will become more obvious once we implement the other compute engines (redis, etc)
The idea is also to have functionalities in some base class or objects.

docling_serve/engines/async_local/worker.py
dolfim-ibm: very good point, I think we will take it into account when doing more refactoring with the other implementations

docling_serve/datamodel/engines.py
dolfim-ibm: will apply styling changes overall when we introduce the new flags

PeterStaar-IBM (dismissed):
good!

cau-git (approved):

cau-git (approved):

PR #64, opened 2025-02-23
These changes speed up python linting step by 10 seconds in total

Fogapod:
There are 3 primary changes:

### pre-commit caching

pre-commit docs have an example of github caching which i applied: https://pre-commit.com/#github-actions-example

### caching python dependencies

`uv sync` dependencies will now be cached by disabling pruning: `prune-cache: false`.
I think you would need to remove `setup-uv-*` caches at https://github.com/DS4SD/docling-serve/actions/caches for this to apply.

### Switching to pre-commit-uv

https://github.com/tox-dev/pre-commit-uv is a patch to pre-commit which uses uv internally which speeds up initial dependency installation. Their caching/configuration is identical, the only change is package installation machinery.
Currently it will be automatically installed as a dev dependency.

I can exclude this patch if this is not desirable


Fogapod:
Uv authors do not recommend caching pip deps as cache will grow over time and parallel downloads are fast. Removed package cache

.github/workflows/job-checks.yml
```diff
@@ -16,10 +16,19 @@ jobs:
...
+        with:
+          path: ~/.cache/pre-commit
+          key: pre-commit|${{ env.PY }}|${{ hashFiles('.pre-commit-config.yaml') }}
+
       - name: Install dependencies
-        run: uv sync --all-extras --no-extra cu124
+        run: uv sync --frozen --all-extras --no-extra cu124
+
       - name: Run styling check
-        run: uv run --no-sync pre-commit run --all-files
```
dolfim-ibm, line 22: `--no-sync` was added because it looked like `uv run` is not respecting the extras set before.

basically it ended up installing all the nvidia drivers from pypi.

.github/workflows/job-checks.yml
```diff
@@ -16,10 +16,19 @@ jobs:
...
+        run: pre-commit run --all-files
```
dolfim-ibm, line 31: It looks like most projects are doing something like:

```
uv tool run --with pre-commit-uv pre-commit run ...
```

this could remove the need for the extra dev deps? or are there recommendations of when using it as dev deps vs as a tool?

Fogapod: I use it locally. Dont know what works for other people

Fogapod: It pulls `uv` inside venv as dev dep which might not be ideal

Fogapod:
CI failed because:
- lint: lock file in repo is outdated
- build: huggingface_hub.errors.HfHubHTTPError: 403 Forbidden: None.

dolfim-ibm:
For the HF timeouts we at the moment we can only retry running.

For the issue in the uv lock file, please update it with the latest main.

Fogapod:
Rebased

dolfim-ibm (approved):

PR #66, opened 2025-02-25
fix: Build arm64 image for docling-serve (gpu)

vishnoianil:
Also fixes the README docling-serve execution instructions.


dolfim-ibm:
This is now deprecated. Images are available in multiple arch whenever possible, see the table in the README.

PR #73, opened 2025-03-03
fix(docs): Remove comma in convert/source curl example

sroecker:
Remove comma in convert/source curl example to make it work.

Otherwise it fails with error message:
{"detail":[{"type":"json_invalid","loc":["body",607],"msg":"JSON decode error","input":{},"ctx":{"error":"Expecting property name enclosed in double quotes"}}]}%

PR #75, opened 2025-03-05
Fix url convert curl example in the readme

nerdalert:
Resolves a json decode err response.


dolfim-ibm:
did you see #73?

nerdalert:
@dolfim-ibm I did not, ty!

PR #78, opened 2025-03-06
feat: display version in fastapi docs

Fogapod:
I don't see version anywhere else in UI currently. I did what `--version` code does

### DIFF
--- a/docling_serve/__main__.py
+++ b/docling_serve/__main__.py
@@ -1,4 +1,4 @@
-import importlib
+import importlib.metadata
 import logging
 import platform
 import sys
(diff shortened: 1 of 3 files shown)


docling_serve/__main__.py
```diff
@@ -1,4 +1,4 @@
-import importlib
+import importlib.metadata
```
Fogapod, line 1: im not sure how it works in docling_serve, it fails in REPL:
```python
>>> import importlib
>>> importlib.metadata.version
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'importlib' has no attribute 'metadata'
```
Also pyright was unhappy about previous import

docling_serve/app.py
```diff
@@ -85,9 +85,17 @@ async def lifespan(app: FastAPI):
 
 
 def create_app():
+    try:
+        version = importlib.metadata.version("docling_serve")
+    except importlib.metadata.PackageNotFoundError:
```
Fogapod, line 90: This shoud be possible if docling_core isn't installed as package for some reason

dolfim-ibm (approved):

PR #95, opened 2025-03-13
chore: move to docling-project gh org

dolfim-ibm:


dolfim-ibm:
I think we still have to rename `name: ds4sd/docling-serve` once the quay.io is available.

PeterStaar-IBM (approved):

PR #104, opened 2025-03-20
fix: Move ARGs to prevent cache invalidation

Fogapod:
According to https://docs.docker.com/reference/dockerfile/#impact-on-build-caching `ARG` casuse cache invalidation.

Overwriting uv args or model list doesn't reinstall OS packages anymore


dolfim-ibm (approved):
lgtm

PR #119, opened 2025-04-03
fix: Fix image permissions to run it with any-user-id on Openshift

### LINKED ISSUE #118: Make image able to run with any-user-id on Openshift
Sometimes we don't control security context of the pod, for example in case of Data Science Pipelines (Kubeflow) and we want to install additional packages on top of existing docling-serve image. Currently it's not possible to write into `/opt/app-root/` with any-user-id and as the result additional modules/apps can't be installed.
Following Red Hat's guidelines on this in "Support arbitrary user ids" section, requires a small change in Containerfile:
[4.1.2. OpenShift Container Platform-specific guidelines](https://docs.redhat.com/en/documentation/openshift_container_platform/4.2/html/images/creating-images#images-create-guide-openshift_create-images)

vku-ibm:
Fixes access to `/opt/app-root/` for any-user-id that is used by Openshift. This allows to install additional packages into image, for example in case of kubeflow pipelines.

Resolves #118

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -52,8 +52,8 @@ RUN echo "Downloading models..." && \
     HF_HUB_DOWNLOAD_TIMEOUT="90" \
     HF_HUB_ETAG_TIMEOUT="90" \
     docling-tools models download -o "${DOCLING_SERVE_ARTIFACTS_PATH}" ${MODELS_LIST} && \
-    chown -R 1001:0 /opt/app-root/src/.cache && \
-    chmod -R g=u /opt/app-root/src/.cache
+    chgrp -R 0 /opt/app-root/ && \
+    chmod -R g=u /opt/app-root/
 
 COPY --chown=1001:0 ./docling_serve ./docling_serve
 RUN --mount=from=ghcr.io/astral-sh/uv:0.6.1,source=/uv,target=/bin/uv \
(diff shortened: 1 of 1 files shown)


Containerfile
```diff
@@ -62,6 +62,9 @@ RUN --mount=from=ghcr.io/astral-sh/uv:0.6.1,source=/uv,target=/bin/uv \
     --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
     uv sync --frozen --no-dev --all-extras ${UV_SYNC_EXTRA_ARGS}
 
+RUN chgrp -R 0 /opt/app-root/ && \
```
dolfim-ibm, line 65: I fear that adding this here will create a huge layer in the container image with a new version of all the files.

I don't a way to verify it now, but can you please check the size of the layers produced by this RUN?

In case, moving it before, would it still work?

vku-ibm: Yes, images size goes up significantly, there is a new layer created that is 7GB. I'll experiment with the placement

vku-ibm: @dolfim-ibm Managed to reduce image size by only 0.5Gb, unfortunately the last block is where it can be done and still working as expected. Perhaps we need to change the way we construct the image from the start.

dolfim-ibm: What is the actual error while installing? Which files/folders are actually problematic in the runtime install?

vku-ibm: Brakes on this error when trying to install python packages:

`ERROR: Could not install packages due to an OSError: [Errno 13] Permission denied: '/opt/app-root/lib/python3.12/site-packages/urllib3-2.3.0.dist-info/'
Check the permissions.`

dolfim-ibm:
Superseded #136

PR #131, opened 2025-04-15
feat(UI): change UI to use async endpoints

SantanaTiago:
UI use async endpoints:

- [x] `task_id` is shown in UI after task is send to queue
- [x] disable output as file as currently async endpoints does not output as file
- [x] rename to accept single file/url
- [x] upload file is converted to base64 string before sending to conversion


dolfim-ibm (approved):
very good!

PR #135, opened 2025-04-16
docs: update deployment examples

rmdg88:
adds deployment examples to leverage the nvidia cuda image

- compose for local deployment
- openshift deployment with service

### DIFF
--- a/docs/deploy-examples/docling-serve-simple.yaml
+++ b/docs/deploy-examples/docling-serve-simple.yaml
@@ -38,6 +38,15 @@ spec:
       restartPolicy: Always
       containers:
         - name: api
+          resources:
+            limits:
+              cpu: 500m
+              memory: 2Gi
+              nvidia.com/gpu: 1  # Limit to one GPU
+            requests:
+              cpu: 250m
+              memory: 1Gi
+              nvidia.com/gpu: 1  # Limit to one GPU
           env:
             - name: DOCLING_SERVE_ENABLE_UI
               value: 'true'
--- a/docs/deploy-examples/docling-serve-simple.yaml
+++ b/docs/deploy-examples/docling-serve-simple.yaml
@@ -1,0 +1,49 @@
+# This example deployment configures Docling Serve with a Service and cuda image
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: docling-serve
+  labels:
+    app: docling-serve
+    component: docling-serve-api
+spec:
+  ports:
+  - name: http
+    port: 5001
+    targetPort: http
+  selector:
+    app: docling-serve
+    component: docling-serve-api
+---
+kind: Deployment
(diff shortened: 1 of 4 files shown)


docs/deploy-examples/docling-serve-simple.yaml
```diff
@@ -0,0 +1,49 @@
...
+      app: docling-serve
+      component: docling-serve-api
+  template:
+    metadata:
+      labels:
+        app: docling-serve
+        component: docling-serve-api
+    spec:
+      restartPolicy: Always
+      containers:
```
dolfim-ibm, line 39: it might be better to add some resources `request` and `limit` for the container.

docs/deploy-examples/docling-serve-simple.yaml
```diff
@@ -0,0 +1,49 @@
...
+        - name: api
+          env:
+            - name: DOCLING_SERVE_ENABLE_UI
+              value: 'true'
+          ports:
+            - name: http
+              containerPort: 5001
+              protocol: TCP
+          imagePullPolicy: Always
+          image: 'ghcr.io/docling-project/docling-serve-cu124'
```
dolfim-ibm, line 49: if we use the GPU image, we should add the limit for requesting the gpu nodes, or not?

docs/deployment.md
```diff
@@ -1,7 +1,152 @@
...
+
+https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#id6
+
+Steps:
+
+1. Check driver version and which GPU you want to use (0/1/2/3.. and update [compose-gpu.yaml](./deploy-examples/compose-gpu.yaml) file or use `count: all`)
+
+```sh
+nvidia-smi
+```
```
dolfim-ibm, line 58: this block should be indented with 4 spaces. similar the other code blocks in the itemization

docs/deployment.md
```diff
@@ -1,7 +1,152 @@
...
+
+[OpenShift deployment](#openshift)
+
+
+_______
+
+## Local GPU deployment
+
+### Docker compose
+
```
dolfim-ibm, line 13: I wonder if we could make use of some collapsed `<details>` section to make it overall more readable. Now the majority of this page is about making CUDA work which is not a really a Docling topic.

dolfim-ibm (commented):
Very good. Let's iterate on a few details and it will be ready

docs/deploy-examples/docling-serve-simple.yaml
rmdg88: added cpu/mem limits

docs/deployment.md
rmdg88: added some details and summary to enclose that extra information

dolfim-ibm (approved):

Issue #103, opened 2025-03-20
Support of SmolDocling


pommedeterresautee:
[SmolDocling](https://huggingface.co/ds4sd/SmolDocling-256M-preview) seems to provide even faster inference and still maintain high quality. Any plan to make it available in the server?

dolfim-ibm:
It will definitely be available soon, but we cannot guarantee it will be faster on all platforms.

pommedeterresautee:
Was thinking to GPU in particular. Right now we are testing multi worker deployment to better use GPU (via unicorn option). Works well to increase GPU usage (it s quite low otherwise), but we are limited by DDR (A10/L40S GPUs). There is a way to move models weights in PyTorch in shared memory so memory foot print is very limited. Too complicated to do with easy OCR + Docling, but with smalldocling single model which does everything may be it s doable. At least something to look at.
Something else we are doing I found useful, is that we split in the client PDF in pages, so each pages is managed in parallel (multi worker / multi instance deployment). May be something else to look at... Or to have a real convert all function (one which batches pages, not a for python loop :) )

oleksii-halahan-work:
@dolfim-ibm are there any updates on this? need to deploy an API for the company I work in, but surprised to see smoldocling is not supported in the API? help will be much appreciated!

archasek:
@pommedeterresautee  how do you split?

pommedeterresautee:
On NodeJs, parse pdf with pdf-lib, create one page pdfs, send in parallel each of them to a pool of docling servers, wait, assemble results in order and enrich with page info (was done before page meta were added). Each docling is set with 4 uvicorn processes. Use A10 GPUs (24g ddr)

FWIW, we had few OOM errors (cuda) not yet sure why, not a specific pdf which raises the issue but servers being on enough days, like a memory leak. Not yet found the cause.

archasek:
@pommedeterresautee great, thanks. why did you decide to split? it was because it performs just better? why not just send 4 full pdfs to your 4 uvicorn workers? :)
btw. i saw a gpu leak reported here https://github.com/docling-project/docling-serve/issues/72

pommedeterresautee:
For better performances (I don t remember exact numbers but it was like 2-3x faster).
When there is one file to process, GPU usage is higher with pages sent independently + multiple processes, I guess because there are step CPU bounds and some other more intense on the GPU.
We process user content and number of files to process varies a lot during the day.

[...]

PR #142, opened 2025-04-21
feat: Expose more conversion options

### LINKED ISSUE #130: How can I update Docling_server to be compatible with Docling 2.30.0
The latest version of the Docling codebase is already 2.30.0. How can I update Docling_server to be compatible with Docling 2.30.0?  Looking forward to your reply amidst your busy schedule.

### LINKED ISSUE #134: Feature request: expose more pipeline options for configuration
Currently only a few pipeline options can be configured: https://github.com/docling-project/docling-serve/blob/5dfb75d3b9a7022d1daad12edbb8ec7bbf9aa264/docling_serve/docling_conversion.py#L112-L120

I don't see a way to set document timeout defined here: https://github.com/docling-project/docling/blob/fa7fc9e63d45f44af57dd6ad7636a2a16f04b8c4/docling/datamodel/pipeline_options.py#L368

Accelerator options like `num_threads` can be set via env vars (`DOCLING_NUM_THREADS`) but `PipelineOptions` does not have env config.

I'm not sure if this should be addressed here or inside docling by adding env config to pipeline class

dolfim-ibm:
- [x] Update all locked dependencies. Refs #130
   - Pydantic 2.11 is currently incompatible with Gradio, so when using the UI we have to pin it
- [x] Use Docling enum and default values in the UI
- [x] Improve default values in the demo UI, e.g. DoclingDocument selected by default
- [x] Expose more options (page range, document timeout, etc). Refs #134 
- ~Expose the choice for the vlm pipeline. Refs #103~ (it will be in its own PR)

Resolves #130
Resolves #134

### DIFF
--- a/docling_serve/datamodel/convert.py
+++ b/docling_serve/datamodel/convert.py
@@ -8,6 +8,7 @@ from docling.datamodel.pipeline_options import (
     EasyOcrOptions,
     PdfBackend,
     TableFormerMode,
+    TableStructureOptions,
 )
 from docling.models.factories import get_ocr_factory
 from docling_core.types.doc import ImageRefMode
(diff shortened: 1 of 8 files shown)


Fogapod:
So all options will have to be duplicated in docling-serve?
Is there a reason options class can't read env same way `AcceleratorOptions` does?

dolfim-ibm:
> So all options will have to be duplicated in docling-serve? Is there a reason options class can't read env same way `AcceleratorOptions` does?

We might have to think more about the `max_num_pages` and `max_file_size`. All the rest I think is already in the right place.

In general, we could think try out new settings in docling-serve, and then move it upstream in docling.

vku-ibm (approved):

PR #158, opened 2025-04-26
docs: fix new default pdf_backend

nkh0472:
according to 

https://github.com/docling-project/docling-serve/blob/37e2e1ad090ef8d2d521183a93756b58f86294e1/docling_serve/datamodel/convert.py#L199-L209


nkh0472:
Also, I want to know is the default timeout value is 20? Does it mean to wait for 20s max for converting?

https://github.com/docling-project/docling-serve/blob/37e2e1ad090ef8d2d521183a93756b58f86294e1/docling_serve/datamodel/convert.py#L100

PR #165, opened 2025-04-30
fix: allow users to set the area threshold for picture descriptions

dolfim-ibm:
From version [v2.30.0](https://github.com/docling-project/docling/releases/tag/v2.30.0) Docling uses a 5% threshold for selecting which images to process. This PR allows users to configure the parameter and use `picture_description_area_threshold=0` to restore the previous behavior.

### DIFF
--- a/docling_serve/datamodel/convert.py
+++ b/docling_serve/datamodel/convert.py
@@ -9,6 +9,7 @@ from docling.datamodel.pipeline_options import (
     EasyOcrOptions,
     PdfBackend,
     PdfPipeline,
+    PictureDescriptionBaseOptions,
     TableFormerMode,
     TableStructureOptions,
 )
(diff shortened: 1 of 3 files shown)


docling_serve/datamodel/convert.py
```diff
@@ -339,6 +340,14 @@ class ConvertDocumentsOptions(BaseModel):
         ),
     ] = False
 
+    picture_description_area_threshold: Annotated[
+        float,
+        Field(
+            description="Minimum percentage of the area for a picture to processed with the models.",
```
ceberam, line 346:
```suggestion
            description="Minimum percentage of the area for a picture to be processed with the models.",
```

ceberam (changes requested):

ceberam (approved):

PR #193, opened 2025-05-18
Adds a cu128 build for Blackwell GPUs support

dawidkulpa:
Hey!

I've added a cu128 build that supports the newest NVIDIA's GPU lineup. I decided to add a new build for it, to keep the support for cu124.
The change was for my own use, but I decided to share it here in case anyone else needs it.
Let me know what you think of this change and if you'd like it merged or dropped :)


dolfim-ibm:
We will soon start reviewing this. We definitely want to support also more Cuda versions.

I was anyway thinking of reviewing the whole conflicting extras issues. Maybe we can find a solution which scales a bit better.

dolfim-ibm:
The new images are now available. See https://github.com/docling-project/docling-serve/pull/229

PR #194, opened 2025-05-18
feat: page break placeholder in markdown exports options 

### LINKED ISSUE #180: Expose page break placeholder in MD
Docling is able to detect and insert a place holder for each page break since v 2.24.0
https://github.com/docling-project/docling-core/releases/tag/v2.24.0

Here the test to call it from code:
https://github.com/docling-project/docling-core/blob/main/test/test_docling_doc.py#L431

Could it be exposed by the server through usage parameters?

Same question for page range, those are quite important parameters.

dolfim-ibm:
Resolves #180


docling_serve/datamodel/convert.py
```diff
@@ -296,6 +296,14 @@ class ConvertDocumentsOptions(BaseModel):
         ),
     ] = 2.0
 
+    md_page_break_placeholder: Annotated[
+        Optional[str],
+        Field(
+            description="Add this placeholder betweek pages in the markdown output.",
+            examples=["<!-- page-break -->", ""],
```
vagenas, line 303: I would remove the `""` from the examples and add `None` instead.

Reason: `""` is implemented in https://github.com/docling-project/docling-core/pull/298 to behave consistently with other strings, which will typically lead to surrounding newlines, which some users may not expect.

ceberam (approved):

vagenas (approved):

PR #197, opened 2025-05-20
docs: Example and instructions on how to load model weights to persistent volume

### LINKED ISSUE #191: Example Kubernetes Job creating a volume with model weights
## Problem description

Models weights are needed for running the pipeline. The current `docling-serve` image has only the simple set of models, but any advanced feature will miss the model weights.

For offline compute, we want to provide a simple way for preparing all the artifacts, which is usually done with the helper CLI `docling-tools models download`.

## Action items

Prepare a set of example manifests for
1. create a persistent volume
2. launch a Kubernetes Jobs using the `docling-serve-cpu` image (we don't need GPU for it) which
    1. mounts the volume
    2. runs the command `docling-tools models download -o <MOUNT-PATH> ${DOCLING_MODELS}`
3. modify the docling-serve Deployment with the volume mount and set `DOCLING_SERVE_ARTIFACTS_PATH` pointing to it.

## Alternatives

We also want to promote and use the OCI artifacts, see https://kubernetes.io/blog/2024/08/16/kubernetes-1-31-image-volume-source/.
However, this is only supported in the latest Kubernetes releases (in alpha version), so we have to provide also a simpler alternative.

vku-ibm:
Changes to documentation:
- instructions on how to download model weights into persistent volume
- example yaml manifests

Resolves #191

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ An easy to use UI is available at the `/ui` endpoint.
 
 ## Documentation and advance usages
 
-Visit the [Docling Serve documentation](./docs/README.md) for learning how to [configure the webserver](./docs/configuration.md), use all the [runtime options](./docs/usage.md) of the API and [deployment examples](./docs/deployment.md).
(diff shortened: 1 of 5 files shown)


dolfim-ibm (approved):
lgtm

PR #203, opened 2025-05-28
fix: properly handle empty complex params 

WiktorSobanski:
Improved handling of the complex parameters, namely `picture_description_local` and `picture_description_api`. 
Sending empty values for these params resulted in 422 validation error, as empty value was parsed to empty string.

Now it handles the params correctly, sending these params can be omitted or sent as empty and the request will be valid.


dolfim-ibm:
Thanks for looking at this (very annoying, I know!) issue.

I was about to also have a deeper look at a solution, more thinking of some field_validator or similar.
What you propose here could be a backup solution, but it is very verbose and redundant. If possible I would like to avoid it.

dolfim-ibm:
Superseded by #210

PR #219, opened 2025-06-11
feat: Enable providing pre-signed PUT-able URL for remote result file storage

trjw:
When hosting docling-serve behind an AWS gateway, there's a hard limit for both request **and response** payloads of 10Mbyte.

Docling's remote fetch is suitable for uploading large docs, however when producing large results (esp json format), the AWS limit gets in the way for any reasonably image-heavy sources.

The simplest way around this is to bounce the result via an S3 or GCS storage bucket, both of which permit the creation of pre-signed, PUT-able URLs for result upload.  Benefit of this approach is that no keys or secrets are required to be available on the docling server - the caller just generates a pre-signed URL and passes it with the request, then fetches the result upon completion.

The proposed patch:
* Adds `remote_storage_url` option to `ConvertDocumentsOptions`.
* Upload conversion result to `remote_storage_url` if provided (via "PUT" method)
* Return `RemoteFileResponse` with URL upon successful upload.
* Force `return_as_file` to true if `remote_storage_url` is active.

### DIFF
--- a/docling_serve/app.py
+++ b/docling_serve/app.py
@@ -5,7 +5,7 @@ import shutil
 import time
 from contextlib import asynccontextmanager
 from io import BytesIO
-from typing import Annotated
+from typing import Annotated, Union
 
 from fastapi import (
     BackgroundTasks,
@@ -43,6 +43,7 @@ from docling_serve.datamodel.responses import (
     ConvertDocumentResponse,
     HealthCheckResponse,
     MessageKind,
+    RemoteFileResponse,
     TaskStatusResponse,
     WebsocketMessage,
 )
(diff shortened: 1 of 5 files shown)


trjw:
In theory it might also be useful to support PUTting the serialised JSON response (instead of zipfile), happy to build that if there's interest.  It's a nicer response format with more job metadata

trjw:
@dolfim-ibm I've revisited this implementation, pre-signed URLs are OK, but not ideal for this.  There's a much cleaner and more generic approach that I'll raise a new PR for, that builds on obstore to provide a generic abstraction for remote storage

dolfim-ibm:
@trjw please note that we are about to integrate the work of https://github.com/docling-project/docling-jobkit in docling-serve. We have then to generalize the source and target to accept s3 coordinates and credentials.

We can sync on the topic if you are interested.

PR #249, opened 2025-06-27
feat!: v1 api with list of sources and target

### LINKED ISSUE #238: Move API to /v1
Drop support for `/v1alpha` and switch to `/v1`.

This change will **not** be backwards compatible.

## Proposed API specs

```yaml
options: {}  # same as the current version, but without the `return_as_file`
sources:
  - kind: file
     base64_string: ""
     filename: ""
  - kind: http
     url: ""
     headers: {}
  - kind: s3  # [coming soon]
     # ...
target:
  kind: inbody  # default (still have to validate the name)
  kind: zip  # equivalent of using return_as_file in the prev versions
  kind: s3  # [coming soon]
  # ...
  kind: put  # [coming soon] PUT zip to a pre-signed url 
```

dolfim-ibm:
Implement the v1 API with list of sources and target. See #238.

TODO:
- [x] Introduce list of sources with `kind`
- [x] Update all endpoints to `/v1`
- [x] Update tests
- [x] Add target
- [x] Use `target` instead of the `return_as_file` option
- [x] What is a good argument for the `/file` endpoints? Maybe keep `return_as_file` only there.
- [x] Update docs
- [x] Pin docling-jobkit release

Resolves #238

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -65,7 +65,6 @@ Simple payload example:
     "pdf_backend": "dlparse_v2",
     "table_mode": "fast",
     "abort_on_error": false,
-    "return_as_file": false,
   },
   "http_sources": [{"url": "https://arxiv.org/pdf/2206.01062"}]
 }
@@ -108,7 +107,6 @@ curl -X 'POST' \
     "pdf_backend": "dlparse_v2",
     "table_mode": "fast",
     "abort_on_error": false,
-    "return_as_file": false,
     "do_table_structure": true,
     "include_images": true,
     "images_scale": 2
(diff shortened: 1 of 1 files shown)


PierreMesure:
It would be great to be able to poll without specifying a task_id and get the list of tasks (not sure how many, that could be specified through a parameter?). I'm using docling-serve to process documents in parallel and I keep polling each task when one request could be enough.

ceberam (approved):

ceberam (approved):

cau-git (approved):

PR #270, opened 2025-07-15
feat: add new source and target

SantanaTiago:
This PR adds:
- new task source `s3`
- new task targets `put` and `s3`

Changes:
- ~~`TargetName` not used in convert file endpoints, switch to `target` with the introduction of `put` target (to be confirmed that it's a valid target option)~~ edit: `put` target to be implemented later

### DIFF
--- a/docling_serve/response_preparation.py
+++ b/docling_serve/response_preparation.py
@@ -251,7 +251,7 @@ def process_results(
         if isinstance(target, PutTarget):
             try:
                 with open(file_path, "rb") as file_data:
-                    r = httpx.put(target.url, files={"file": file_data})
+                    r = httpx.put(str(target.url), files={"file": file_data})
                     r.raise_for_status()
                 response = PresignedUrlConvertDocumentResponse(
                     status=conv_result,


docling_serve/response_preparation.py
```diff
@@ -234,9 +246,40 @@ def process_results(
         # Output directory
         # background_tasks.add_task(shutil.rmtree, work_dir, ignore_errors=True)
 
-        response = FileResponse(
-            file_path, filename=file_path.name, media_type="application/zip"
-        )
+        if isinstance(target, PutTarget):
```
dolfim-ibm, line 249: `PutTarget` should already be a presigned URL provided by the user. Then in the block below we should simply use `httpx.put`

docling_serve/datamodel/requests.py
```diff
@@ -19,15 +28,22 @@ class HttpSourceRequest(HttpSource):
...
+    kind: Literal["s3"] = "s3"
+
+
 ## Multipart targets
+## not used with put target
 class TargetName(str, enum.Enum):
     INBODY = InBodyTarget().kind
     ZIP = ZipTarget().kind
+    # S3 = S3Target().kind
+    # PUT = PutTarget().kind
```
dolfim-ibm, line 41: PUT could indeed be used in the file endpoint, but it can be a new PR

docling_serve/datamodel/responses.py
```diff
@@ -35,6 +35,10 @@ class ConvertDocumentResponse(BaseModel):
     timings: dict[str, ProfilingItem] = {}
 
 
+class PresignUrlConvertDocumentResponse(BaseModel):
```
dolfim-ibm, line 38:
```suggestion
class PresignedUrlConvertDocumentResponse(BaseModel):
```

prein:
Thanks for adding s3 source! I'm trying it out from your fork right now. Should it work with pod identity or IRSA? Or only access/secret key?

docling_serve/datamodel/responses.py
```diff
@@ -35,6 +35,10 @@ class ConvertDocumentResponse(BaseModel):
     timings: dict[str, ProfilingItem] = {}
 
 
+class PresignedUrlConvertDocumentResponse(BaseModel):
+    url: str
+
```
dolfim-ibm, line 40: let's keep `status` and `processing_time` from the other type of response (`ConvertDocumentResponse`)

docling_serve/datamodel/responses.py
```diff
@@ -35,6 +35,10 @@ class ConvertDocumentResponse(BaseModel):
     timings: dict[str, ProfilingItem] = {}
 
 
+class PresignedUrlConvertDocumentResponse(BaseModel):
+    url: str
```
dolfim-ibm, line 39: `url` is a bit wrong here, because we don't have the GET url. I think we better skip it.

dolfim-ibm (dismissed):
nice

dolfim-ibm (approved):
lgtm, nice

PR #293, opened 2025-07-31
fix: null type in downgraded openapi schema

SantanaTiago:
This PR fixes `null` in `anyof` schemas to be compatible with 3.0.x


dolfim-ibm:
no longer needed

Issue #288, opened 2025-07-29
Split processing example


dolfim-ibm:
## Problem

Long documents (many pages) will take a linear amount of time when processing. Could we leverage splitting the file in multiple sub-documents and process them in parallel?

## Proposal

A client code using Docling Serve could leverage multiple requests on different page ranges to perform those conversions in parallel. Then, parsing the json response with the `DoclingDocument` in docling-core, we should be able to merge the the page ranges.

Let's make some example code which could be shared when users get into this issue.

Fogapod:
So its possible to do this without losing accuracy like some multi page tables getting broken?

PR #303, opened 2025-08-07
docs: add split processing example

SantanaTiago:
Resolves [#288]


dolfim-ibm:
@dosu propose a documentation page describing this new example.

PR #310, opened 2025-08-08
ci: workflow improvements

rmdg88:
regarding build and tests:
- updated actions versions
- the new test stage, runs after the build, and so it uses the build cache, one is just exporting the build cache to docker 
```load: true # == '--output=type=docker'```
- it runs for all images
- stage does: mounts container, checks health, runs pytest, cleanup
- adds images tests for released, `-cpu` and `-cu*`  (commented out now,maybe not needed anymore)

regarding releases:
- adds docling packages versions to changelog, from ```uv.lock```

ref https://github.com/docling-project/docling-serve/issues/302

### DIFF
--- a/.github/workflows/job-checks.yml
+++ b/.github/workflows/job-checks.yml
@@ -10,7 +10,7 @@ jobs:
       matrix:
         python-version: ['3.12']
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - name: Install uv and set the python version
         uses: astral-sh/setup-uv@v6
         with:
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: .github/workflows/job-checks.yml [hunk 1]
```diff
@@ -33,36 +33,38 @@ jobs:
...
 
-  markdown-lint:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-      - name: markdownlint-cli2-action
-        uses: DavidAnson/markdownlint-cli2-action@v16
-        with:
-          globs: "**/*.md"
+  # markdown-lint:
```
dolfim-ibm, line 63: what is the issue here?

dolfim-ibm (approved):
lgtm

PR #343, opened 2025-09-01
docs: add metrics deployment example

rmdg88:
- adds:
  -  servicemonitor resource example for prometheus
  -  metrics documentation and examples

- improves:
  - rq deployment example by adding securityContext, so one can deploy in openshift


dolfim-ibm:
Addressed in the OTEL deployment

PR #358, opened 2025-09-09
feat: add OpenTelemetry with OTLP exporter

### LINKED ISSUE #357: Support OpenTelemetry with OTLP exporter
Enable opentelemetry and OTLP exporter so that requests and responses are visible within a wider system, open telemetry integration would be great to have out of the box with standard environment variables for configuration.

fastapi has automatic instrumentation support from the openetlemetry-python-contrib repo and is available here https://pypi.org/project/opentelemetry-instrumentation-fastapi/

If the environment variables aren't configured, then no exporting of traces would occur as expected

damienpontifex:
- Resource with default service name
- OTLP exporter that will use standard environment variables for configuration
- FastAPI auto instrumentation for app
- Adds a docker compose setup for validating the tracing with `docker compose up -d --build` and Grafana LGTM stack for viewing/inspecting the traces. Screenshot below
image

Resolves #357


dolfim-ibm:
@damienpontifex Otel could be used also for telemetry such as response time dashboard, etc, right? do you have any example how this would look like?

damienpontifex:
@dolfim-ibm 💯 
In the example/screenshot above, I'm using Grafana, so to follow on with that, here's a good overview of a lot of things you can do with Otel traces/metrics out of the box https://grafana.com/products/cloud/application-observability/
This would be valid for any Otel backend e.g. I've also used azure app insights, dynatrace etc etc, it's all capable of doing the washboarding like you say using Otel data

prein:
@damienpontifex , does this add traces and metrics or only traces? It looks to me that it's traces only, which may be helpful to clarify.

mauricionr:
looks like a lint fail here, I made a pr for that, https://github.com/damienpontifex/docling-serve/pull/1 @damienpontifex do you think that my pr can help here and we shall continue with this feat rollout?

PR #382, opened 2025-09-25
fix: Correctly raise HTTPException for Gateway Timeout

### LINKED ISSUE #380: HTTPException not correctly raised, leading to ResponseValidationError on Gateway Timeout
## Description

When a synchronous document conversion task exceeds the DOCLING_SERVE_MAX_SYNC_WAIT limit, the API endpoints (e.g., /v1/convert/source) currently **return** HTTPException(...) instead of **raise** HTTPException(...).

This behavior causes FastAPI to attempt to serialize the HTTPException instance against the declared response_model (e.g., ConvertDocumentResponse or ChunkDocumentResponse), which expects fields like document, status, and processing_time. Consequently, a fastapi.exceptions.ResponseValidationError is raised, indicating "Field required" errors for these missing fields in the HTTPException object, rather than returning the intended 504 Gateway Timeout HTTP response.

## How to reproduce

Make a conversion request using a sync endpoint, e.g. /v1/convert/source, that takes longer time than `DOCLING_SERVE_MAX_SYNC_WAIT`. This env variable defaults to 120 seconds. You can set a lower value to more easily reproduce the problem, like `DOCLING_SERVE_MAX_SYNC_WAIT=5` when starting Docling Serve.

### Example using Docker
`docker run -p 5001:5001 -e DOCLING_SERVE_MAX_SYNC_WAIT=5 quay.io/docling-project/docling-serve-cu124:latest`

erikmargaronis:
Correctly raise HTTPException for Gateway Timeout instead of returning it.

Resolves #380

### DIFF
--- a/docling_serve/app.py
+++ b/docling_serve/app.py
@@ -455,7 +455,7 @@ def create_app():  # noqa: C901
 
         if not completed:
             # TODO: abort task!
-            return HTTPException(
+            raise HTTPException(
                 status_code=504,
                 detail=f"Conversion is taking too long. The maximum wait time is configure as DOCLING_SERVE_MAX_SYNC_WAIT={docling_serve_settings.max_sync_wait}.",
             )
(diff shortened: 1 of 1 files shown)


dolfim-ibm (approved):
lgtm

Issue #387, opened 2025-10-01
UVICORN_ROOT_PATH doesn't work with Gradio UI


bbeiler-ridgeline:
Running this locally via docker, I have the following setup
```
    environment:
      - UVICORN_PORT=5001
      - UVICORN_WORKERS=1
      - DOCLING_SERVE_ENG_LOC_NUM_WORKERS=2
      - UVICORN_ROOT_PATH=/docling
      - DOCLING_SERVE_ENABLE_REMOTE_SERVICES=true
      - DOCLING_SERVE_ENG_LOC_SHARE_MODELS=true
      - DOCLING_SERVE_ENABLE_UI=True
      - DOCLING_SERVE_MAX_SYNC_WAIT=360
      - DOCLING_SERVE_LOAD_MODELS_AT_BOOT = true
```
And I use an nginx reverse proxy that routes /docling calls to this container. The API works great, but the gradio UI is unreachable, either by direct `5001/ui` access or by the expected root path `/docling/ui`.

I ran a deepwiki check and it looks like maybe the gradio mounting doesn't take into account the root path arg?
```
            app = gr.mount_gradio_app(
                app,
                gradio_ui,
                path="/ui",
                allowed_paths=["./logo.png", tmp_output_dir],
                root_path="/ui",
            )
```
https://deepwiki.com/search/when-i-use-uvicornrootpath-to_440bbb39-cfbf-4526-80f5-8fbad5c7ef25

bbeiler-ridgeline:
Yah, I think the mention of `use the value from the UVICORN_ROOT_PATH environment variable when calling mount_gradio_app` would be a great add. Will look at putting up an MR for that if possible, or unless someone else is able to slip that in with an adjacent MR

sahlex:
Upvote! I have the same issue.

Regards

PR #396, opened 2025-10-08
fix: run docling ui behind a reverse proxy using a context path 

sahlex:
Setting 
`UVICORN_ROOT_PATH=/docling `
now allows to run docling ui behind a reverse proxy.

This addresses  https://github.com/docling-project/docling-serve/issues/387


dolfim-ibm (changes requested):
The root_path change looks good, but please don't push any of the other files if they are not needed.

bbeiler-ridgeline:
@sahlex Thanks for pushing this up! Was just working on a fork/PR when I noticed you had already got the fix going!

PR #495, opened 2026-02-12
chore: sync upstream v1.12.0 and fix Docker build

lbadams3:
## Summary
- Sync fork with upstream docling-serve v1.12.0 (from v1.9.0), bringing in OTel tracing, metrics endpoint, and updated dependencies
- Use `uv sync --frozen` with `uv.lock` to pin dependency versions, preventing version drift that caused `libxcb.so.1` runtime errors
- Replace `opencv-python` with `opencv-python-headless` since the FIPS base image has no X11 libraries
- Remove `picture_classifier` from downloaded models
- Deduplicate `serve` COPY into a single layer with correct ownership and permissions

## Test plan
- [x] Docker image builds successfully with `--platform linux/amd64`
- [x] Container starts without import errors (`cv2`, `libxcb`, etc.)
- [x] Async document conversion via `/invocations` endpoint completes successfully
- [x] PDF processing returns expected response keys (`document`, `status`, `errors`, `processing_time`, `timings`)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,20 +3,21 @@ ENV UV_COMPILE_BYTECODE=0 UV_LINK_MODE=copy UV_PYTHON_DOWNLOADS=0
 
 WORKDIR /app
 
-COPY ./pyproject.toml ./pyproject.toml
+COPY ./pyproject.toml ./uv.lock ./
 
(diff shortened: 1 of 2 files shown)


lbadams3:
Opened against wrong repo by mistake, sorry.

PR #498, opened 2026-02-16
fix: prevent WebsocketNotifier crash when task has no subscribers

### LINKED ISSUE #497: Unhandled RuntimeError in WebsocketNotifier crashes pubsub loop and causes pod CrashLoopBackOff
### Description

  When a task update arrives via Redis pubsub for a task that has no websocket
  subscribers (e.g. the client is using HTTP polling via  GET
  /v1/status/poll/{task_id} ),  WebsocketNotifier.notify_task_subscribers()
  raises a  RuntimeError . This exception is not caught by the upstream
  _listen_for_updates()  loop in  docling-jobkit , which kills  process_queue() ,
  tears down the FastAPI lifespan, and crashes the pod.

  On Kubernetes, this causes a crash loop: the pod restarts, the same task update
  is still in the Redis pubsub backlog, the same error fires again, and the pod
  enters  CrashLoopBackOff  after ~4 restarts.

  ### Root Cause

  Two issues:

  1.  websocket_notifier.py:30-31  —  notify_task_subscribers()  raises
  RuntimeError  when  task_id  is not in  self.task_subscribers . This is expected
  for any task whose client uses HTTP polling instead of websockets — no
  subscriber set is ever created for those tasks.
  2.  app.py:925  — The websocket endpoint accesses  orchestrator.notifier.
  task_subscribers[task_id]  directly without initializing the set first, which
  would raise  KeyError  if the entry didn't happen to exist.

  ### Reproduction

  1. Deploy docling-serve with the RQ async engine
  2. Submit a document via  POST /v1/convert/file/async
  3. Poll for status via  GET /v1/status/poll/{task_id}  (do NOT connect via
  websocket)
  4. When the RQ worker publishes a task update, the notifier raises  RuntimeError
  5. Pod crashes

  ### Error Log

    RuntimeError: Task c065a021-c605-4819-b5ee-4e9983b00171 does not have a
  subscribers list.
      File "docling_serve/websocket_notifier.py", line 31, in
  notify_task_subscribers
      File "docling_jobkit/orchestrators/rq/orchestrator.py", line 209, in
  _listen_for_updates

prein:
**Fix**

•  websocket_notifier.py  — Return silently when  task_id  has no subscribers instead of raising
•  app.py  — Use  setdefault(task_id, set())  when adding a websocket subscriber

Resolves #497

PS. I'm running this in production for some time already

### DIFF
--- a/docling_serve/websocket_notifier.py
+++ b/docling_serve/websocket_notifier.py
@@ -1,3 +1,5 @@
+import logging
+
 from fastapi import WebSocket
 
 from docling_jobkit.datamodel.task_meta import TaskStatus
@@ -9,6 +11,8 @@ from docling_serve.datamodel.responses import (
     TaskStatusResponse,
     WebsocketMessage,
 )
+
+_log = logging.getLogger(__name__)
 
 
 class WebsocketNotifier(BaseNotifier):
@@ -28,8 +32,9 @@ class WebsocketNotifier(BaseNotifier):
 
     async def notify_task_subscribers(self, task_id: str):
         if task_id not in self.task_subscribers:
-            # Task has no websocket subscribers - this is normal for tasks that were
-            # never subscribed to via websocket, or during shutdown cleanup
+            _log.debug(
+                f"Task {task_id} has no websocket subscribers, skipping notification."
+            )
             return
 
         try:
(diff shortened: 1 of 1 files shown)


dolfim-ibm:
@prein we should be able to finalize the docling-jobkit release tomorrow, then we will include all in docling-serve.

Issue #481, opened 2026-01-30
RQ jobs disappearing but task is reported as pending


bjornbugge:
I have docling-serve with the rq orchestrator using redis and with one API pod and n (usually n=1) worker pods.  Our application submits jobs to the docling-serve-api pod and polls the resulting task id regularly until it's reported as either completed, failed, or not found.  In the logs for the docling-serve-api container I see many lines like the snippet below repeated over and over again
```
ERROR:docling_serve.orchestrator_factory:RQ check f8c3a37a-23c0-46bc-9f44-40b1f5cdbeba: No such job: rq:job:f8c3a37a-23c0-46bc-9f44-40b1f5cdbeba
ERROR:docling_serve.orchestrator_factory:RQ check f8c3a37a-23c0-46bc-9f44-40b1f5cdbeba: No such job: rq:job:f8c3a37a-23c0-46bc-9f44-40b1f5cdbeba
ERROR:docling_jobkit.orchestrators.rq.orchestrator:An error occour getting queue position.
Traceback (most recent call last):
  File "/opt/app-root/lib64/python3.12/site-packages/docling_jobkit/orchestrators/rq/orchestrator.py", line 160, in get_queue_position
    job = Job.fetch(task_id, connection=self._redis_conn)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.12/site-packages/rq/job.py", line 654, in fetch
    job.refresh()
  File "/opt/app-root/lib64/python3.12/site-packages/rq/job.py", line 1013, in refresh
    raise NoSuchJobError(f'No such job: {self.key}')
rq.exceptions.NoSuchJobError: No such job: rq:job:f8c3a37a-23c0-46bc-9f44-40b1f5cdbeba
INFO:     100.64.9.78:40452 - "GET /v1/status/poll/f8c3a37a-23c0-46bc-9f44-40b1f5cdbeba HTTP/1.1" 200 OK
```

Looking at the queues in some of our dev deployments that perform this polling I can see that the jobs in question have been all been inserted as part of our internal nightly deployments where several documents are submitted.  I tried to add redisinsight to our K8s cluster and I can see that the UUIDs mentioned in the errors are all present under `docling:tasks:*`, but indeed not under `rq:job:*`.

It seems like the issue is in the call to `task_status_poll`
```python
        try:
[...]

PR #502, opened 2026-02-19
feat: Surface task error messages in status API responses

prein:
### Problem

When a docling task fails, the consumer polling  `/v1/status/poll/{task_id}` receives:

    {"task_id": "...", "task_status": "failure", "error_message": null}

The error reason is logged on the worker and recorded in the OTEL span, but never reaches the API response. This makes it impossible to:
  • Build Grafana dashboards that break down failures by error type
  • Debug failed tasks without access to worker logs
  • Distinguish between timeout, OOM, corrupt PDF, and other failure modes

### Changes

Depends on: [docling-jobkit  feat/error-message-propagation  branch.](https://github.com/docling-project/docling-jobkit/pull/98)

1.  TaskStatusResponse  — add  `error_message: Optional[str] = None`
  •  docling_serve/datamodel/responses.py
2.  rq_job_wrapper.py  — pass  `error_message=str(e)` in the failure  _TaskUpdate  publish
  • Also includes diagnostic logging improvements: source-level context (source_info  list), file hashing for FileSource, and enriched error logging with source context.
3.  orchestrator_factory.py  — store and retrieve  error_message  in Redis task metadata
  •  _store_task_in_redis()  includes  error_message  in the JSON payload
  •  _get_task_from_redis()  reads  error_message  back (defaults to  None for old records)
4.  app.py  — all 7  TaskStatusResponse  constructions now pass error_message=task.error_message
  •  process_source_async ,  process_file_async
  •  chunk_source_async ,  chunk_file_async
  •  task_status_poll
  •  task_status_ws  (connection + update messages)

### API Change (backward-compatible)

Before:

    {"task_id": "abc", "task_type": "convert", "task_status": "failure",  "task_position": null, "task_meta": null}

After:

    {"task_id": "abc", "task_type": "convert", "task_status": "failure",  "task_position": null, "task_meta": null, error_message": "RuntimeError: No converter"}

The  `error_message`  field is `null` for successful tasks and for legacy records without the field.

### Testing

7 new tests in  tests/test_error_message_propagation.py :

•  TaskStatusResponse  model: field presence, JSON output, backward compatibility
• Redis storage: store-and-retrieve roundtrip with error_message, backward compat with old records missing the field

**Issue resolved by this Pull Request:**
Resolves #501

PR #516, opened 2026-02-24
fix: zombie task cleanup — reconcile stale RQ/Redis state

### LINKED ISSUE #515: Zombie task loop: orchestrator polls indefinitely when RQ job is gone but Redis says SUCCESS
## Bug description

After a Redis restart (or RQ job TTL expiry), docling-serve enters an infinite polling loop for completed tasks. 
The task finished successfully but was never removed from the orchestrator's tracking.
Each subsequent status check triggers the same RQ→Redis→RQ cycle with no backoff, no circuit breaker, and no cleanup.

Observed running for 8+ hours on a single pod with no resolution.

Image

**TL;DR** The loop is caused by a **Redis ↔ RQ state mismatch**:  

- Redis says the task is `SUCCESS`  
- RQ says `No such job` (job already cleaned up)  
- The orchestrator polls both every ~10-13 seconds, gets conflicting answers, and just... keeps going

The 3-step cycle repeating every ~10s:  
Status check → RQ lookup → `No such job` → Redis → `SUCCESS` → repeat 

## Reproduction

1. Submit a conversion task via  /v1/convert/source  (async)
2. Task completes successfully (RQ job finishes, pub/sub fires  SUCCESS )
3. Restart Redis, or wait for RQ's  result_ttl  (4h) to expire the job
4. Consumer polls  /v1/status/poll/{task_id}  (or any internal status check fires)
5. The following 3-step cycle repeats every ~10 seconds indefinitely:
  Step 1:  `task_status() → _get_task_from_rq_direct(task_id)`
  Step 2:  `Job.fetch(task_id) → NoSuchJobError → caught by except Exception → returns None`
  Step 3:  `_get_task_from_redis(task_id) → finds cached metadata → returns Task(status=SUCCESS) → consumer gets SUCCESS but result is gone → keeps polling`

## Root cause analysis

The `RedisTaskStatusMixin.task_status()` method
(`orchestrator_factory.py:36-93`) has a 3-tier fallback chain:

```
1. RQ direct check  (_get_task_from_rq_direct)
2. Redis cache       (_get_task_from_redis)
3. Parent in-memory  (super().task_status)
```

When the RQ job no longer exists:

| Tier | Result | Problem |
|------|--------|---------|
| RQ direct | `None` (NoSuchJobError caught) | Returns `None`, falls through |
| Redis cache | `Task(status=SUCCESS)` with 24h TTL | Returns stale success status |
| In-memory | May also have stale entry in `self.tasks` | Never cleaned up |

The method returns `SUCCESS` from Redis cache, but the **result data** (msgpack
blob stored under `docling:results:*`) is already gone - its TTL
(`results_ttl=4h`) is shorter than the metadata TTL (`ex=86400`=24h). So:

- Status says `SUCCESS`
- `/v1/result/{task_id}` returns 404 ("Task result not found")
- Consumer keeps polling, hoping the result will appear
- docling-serve keeps doing the full RQ→Redis lookup cycle every request

### Why the task is never cleaned up

1. **`self.tasks` dict grows unboundedly** - completed tasks are only removed
   by `delete_task()` (explicit) or `clear_results()` (time-based, but only
   called by `/v1/clear/results` endpoint - never automatically)

2. **`single_use_results`** (the only automatic cleanup) is off by default

3. **No zombie detection** - there is no logic that says "if RQ reports
   NoSuchJobError for a task we think is complete, remove it from tracking"

4. **Redis metadata outlives result data** - metadata TTL (24h) > result TTL
   (4h), creating a window where status is `SUCCESS` but result is gone

5. **`_get_task_from_rq_direct` swallows the signal** - `NoSuchJobError` is
   caught by `except Exception` and returns `None`, indistinguishable from
   "task doesn't exist in RQ yet"

### The stale-status verification only makes it worse

The "CRITICAL FIX" block at lines 62-79 rechecks RQ for `PENDING`/`STARTED`
tasks, but **not for `SUCCESS`/`FAILURE`**. This means:

- A `SUCCESS` task with a gone RQ job is returned as-is from Redis
- A `PENDING` task with a gone RQ job gets rechecked (RQ returns `None`), so
  the stale PENDING is preserved - also a zombie, just less visible

## Impact

- **Log noise**: thousands of redundant RQ+Redis lookups per zombie task
- **Redis load**: each poll does 2 Redis operations (RQ fetch + metadata get)
- **Memory leak**: `self.tasks` dict never shrinks for completed tasks
- **Consumer confusion**: gets `SUCCESS` status but 404 on result fetch
- **Pod resource waste**: observed running 8+ hours at ~10s intervals on
  `docling-serve-748879b67-h7lbc`

## Expected behavior

When a task's RQ job no longer exists:

1. If status is `SUCCESS`/`FAILURE` and result is gone → mark task as
   terminal, stop tracking, return final status with a flag indicating
   result has expired
2. If status is `PENDING`/`STARTED` and RQ job is gone → mark as `FAILURE`
   (orphaned task), store error message explaining the job was lost
3. In all cases: remove from `self.tasks` after a reasonable grace period
   (or immediately for terminal+result-expired)

prein:
## Summary

- **Fix infinite polling loop** when RQ jobs expire but Redis metadata persists, causing 8+ hour zombie cycles per task
- **Distinguish "job definitively gone" from "transient RQ error"** by catching `NoSuchJobError` separately and returning a `_RQJobGone` sentinel
- **Reconcile conflicting state** in `task_status()`: terminal tasks get cleaned up, orphaned non-terminal tasks get marked `FAILURE` with a descriptive `error_message`
- **Add background zombie reaper** that periodically sweeps completed tasks from in-memory tracking (configurable interval/max-age)
- **Align TTLs** so Redis metadata and result keys use `eng_rq_results_ttl` (4h) instead of hardcoded 86400 (24h), closing the 20h window where status says SUCCESS but result is 404
- **Surface `error_message`** on `TaskStatusResponse` so consumers can see why a task failed

## Motivation

After Redis restart or RQ job TTL expiry, the orchestrator enters an infinite poll loop: RQ says "no such job" (caught by a generic `except Exception`), Redis says SUCCESS, and nothing reconciles the conflict. Observed running 8+ hours on production pods. Consumers get `SUCCESS` status but 404 on result fetch, and keep polling indefinitely.

Root causes:
1. `_get_task_from_rq_direct` swallowed `NoSuchJobError` — indistinguishable from "RQ temporarily unavailable"
2. `task_status()` never handled the "job is definitively gone" case for terminal statuses
3. `self.tasks` dict grew unboundedly — completed tasks were never cleaned up automatically
4. Metadata TTL (24h) outlived result TTL (4h), creating stale SUCCESS entries

### `docling_serve/orchestrator_factory.py`
- **Layer A**: New `_RQJobGone` sentinel class; `_get_task_from_rq_direct` now catches `NoSuchJobError` separately and returns `_RQ_JOB_GONE` instead of `None`
- **Layer B**: Rewritten `task_status()` with zombie reconciliation:
  - RQ gone + Redis terminal → return status, clean up `self.tasks` and `_task_result_keys`
  - RQ gone + Redis non-terminal → mark `FAILURE` with descriptive `error_message`, clean up tracking
  - RQ gone + not in Redis → raise `TaskNotFoundError`
  - RQ transient error → fall through to Redis/parent as before
- **Layer C**: New `_reap_zombie_tasks()` async method — periodic background sweep of `self.tasks` for completed entries older than `max_age`
- **Layer E**: `_store_task_in_redis` and result key storage now use `eng_rq_results_ttl` instead of hardcoded `86400`
- `_get_task_from_redis` now reads `error_message` from stored metadata
- `_store_task_in_redis` now persists `error_message` when present

### `docling_serve/settings.py`
- **Layer D**: New settings `zombie_reaper_interval` (default 300s) and `zombie_reaper_max_age` (default 3600s)

### `docling_serve/app.py`
- Wire zombie reaper into lifespan: launched as `asyncio.create_task` alongside `process_queue`, cancelled on shutdown

### `docling_serve/datamodel/responses.py`
- Add `error_message: Optional[str] = None` to `TaskStatusResponse`

### `docling_serve/app.py` (poll endpoint)
- Pass `error_message` through in `task_status_poll` response

## No docling-jobkit changes needed

The required fields (`error_message`, `is_completed()`, `set_status()`, `finished_at`) already exist on the `Task` model in docling-jobkit. The `NoSuchJobError` comes from the `rq` package (already a dependency).

## Test plan

- [x] 16 new unit tests in `tests/test_zombie_task_cleanup.py`:
  - `TestRQJobGoneSentinel` (3 tests): sentinel on NoSuchJobError, None on generic error, Task on live job
  - `TestTaskStatusReconciliation` (7 tests): all zombie state combinations (gone+success, gone+failure, gone+pending, gone+started, gone+no-redis, transient+redis, live-rq)
  - `TestZombieReaper` (4 tests): reaps old completed, keeps recent, keeps in-progress, reaps old failed
  - `TestTTLAlignment` (1 test): metadata TTL uses `eng_rq_results_ttl`
  - `TestErrorMessagePropagation` (1 test): error_message round-trips through Redis
- [x] All 16 pass, existing tests unaffected
- [x] **Live test**: deploy Docker image, submit tasks, then simulate:
  1. Delete RQ job manually → poll should return final status + cleanup
  2. Restart Redis → orphaned PENDING tasks should transition to FAILURE with error_message
  3. Wait for reaper interval → old completed tasks should be removed from memory
  4. Verify promptservice sees `error_message` in poll response

## Configuration

| Setting | Default | Description |
|---------|---------|-------------|
| `DOCLING_SERVE_ZOMBIE_REAPER_INTERVAL` | `300.0` | Seconds between zombie reaper sweeps |
| `DOCLING_SERVE_ZOMBIE_REAPER_MAX_AGE` | `3600.0` | Remove completed tasks older than this from memory |

Resolves #515 #481

### DIFF
--- a/docling_serve/orchestrator_factory.py
+++ b/docling_serve/orchestrator_factory.py
@@ -113,11 +113,12 @@ class RedisTaskStatusMixin:
                         f"— marking as FAILURE (orphaned)"
                     )
                     task.set_status(TaskStatus.FAILURE)
-                    task.error_message = (
-                        f"Task orphaned: RQ job expired while status was "
-                        f"{task.task_status}. Likely caused by worker restart or "
-                        f"Redis eviction."
-                    )
+                    if hasattr(task, "error_message"):
+                        task.error_message = (
+                            f"Task orphaned: RQ job expired while status was "
+                            f"{task.task_status}. Likely caused by worker restart or "
+                            f"Redis eviction."
+                        )
                     self.tasks.pop(task_id, None)
                     self._task_result_keys.pop(task_id, None)
                     await self._store_task_in_redis(task)
@@ -169,15 +170,15 @@ class RedisTaskStatusMixin:
                 meta.setdefault("num_succeeded", 0)
                 meta.setdefault("num_failed", 0)
 
-                task = Task(
-                    task_id=data["task_id"],
-                    task_type=data["task_type"],
-                    task_status=TaskStatus(data["task_status"]),
-                    processing_meta=meta,
-                    error_message=data.get("error_message"),
-                )
-                if data.get("error_message"):
-                    task.error_message = data["error_message"]
+                task_kwargs: dict[str, Any] = {
+                    "task_id": data["task_id"],
+                    "task_type": data["task_type"],
+                    "task_status": TaskStatus(data["task_status"]),
+                    "processing_meta": meta,
+                }
+                if data.get("error_message") and "error_message" in Task.model_fields:
+                    task_kwargs["error_message"] = data["error_message"]
+                task = Task(**task_kwargs)
                 return task
         except Exception as e:
             _log.error(f"Redis get task {task_id}: {e}")
@@ -282,8 +283,6 @@ class RedisTaskStatusMixin:
                 "processing_meta": meta,
                 "error_message": getattr(task, "error_message", None),
             }
-            if task.error_message:
-                data["error_message"] = task.error_message
 
             metadata_ttl = docling_serve_settings.eng_rq_results_ttl
             async with redis.Redis(connection_pool=self._redis_pool) as r:


prein:
had to rebase on main and force push as the diff was showing changes from https://github.com/docling-project/docling-serve/pull/502 which are already merged

prein:
Chart showing the volume of "No such job" errors from the moment Redis was restarted until this fix was deployed. And another one to zoom in to deployment.
It was already cleaning up on it's own, but you can still see the impact 
image
image

PR #534, opened 2026-03-05
fix: drop otel metrics for excluded endpoints

### LINKED ISSUE #533: OTEL health/metrics endpoints excluded from traces but not from metrics
## Description

The `HealthMetricsFilterSampler` introduced in #456 correctly drops **trace spans** for `/health`, `/metrics`, `/healthz`, `/readyz`, and `/livez`. However, the OTEL **HTTP metrics** (duration histograms, active request counters) are still recorded for these endpoints.

This is because the custom sampler only controls span sampling — the `OpenTelemetryMiddleware` ASGI layer records metrics in a `finally` block that runs regardless of whether the span was sampled or dropped. In production, this creates significant metric noise from Kubernetes health probes (every 5–10s per pod × dozens of pods) and Prometheus scrapes of `/metrics` (every 30s).

## Root cause

In [`otel_instrumentation.py`](https://github.com/docling-project/docling-serve/blob/main/docling_serve/otel_instrumentation.py), `FastAPIInstrumentor.instrument_app(app)` is called without an `excluded_urls` parameter. The `excluded_urls` mechanism in `opentelemetry-instrumentation-fastapi` short-circuits at the ASGI middleware level **before** any span or metric recording occurs, while the custom sampler only affects span creation:

| Filtering approach | Traces filtered? | Metrics filtered? |
|---|---|---|
| `HealthMetricsFilterSampler` (current) | Yes | **No** |
| `excluded_urls` parameter (proposed) | Yes | Yes |

The relevant code path in `opentelemetry-instrumentation-asgi` `OpenTelemetryMiddleware.__call__`:

```python
# excluded_urls check — returns early, skipping ALL instrumentation
if self.excluded_urls and self.excluded_urls.url_disabled(url):
    return await self.app(scope, receive, send)

# ... span creation (controlled by sampler) ...

# finally block — records metrics REGARDLESS of sampling decision
finally:
    duration_histogram_old.record(...)  # ← still fires for dropped spans
    active_requests_counter.add(-1, ...)
```

## Proposed fix

Pass `excluded_urls` derived from `FILTERED_PATHS` to `FastAPIInstrumentor.instrument_app()`:

```python
# Before
FastAPIInstrumentor.instrument_app(app)

# After
excluded_urls = ",".join(f"{p}$" for p in FILTERED_PATHS)
FastAPIInstrumentor.instrument_app(app, excluded_urls=excluded_urls)
```

The `$`-anchored regexes prevent false matches on paths like `/healthcheck` while correctly matching `/health`, `/health?verbose=true`, etc.

The existing `HealthMetricsFilterSampler` can remain as a defense-in-depth layer for trace filtering.

## Workaround

Deployments can set the `OTEL_PYTHON_FASTAPI_EXCLUDED_URLS` environment variable to achieve the same effect without a code change:

```
OTEL_PYTHON_FASTAPI_EXCLUDED_URLS=/health$,/metrics$,/healthz$,/readyz$,/livez$
```

## Impact

In a Kubernetes deployment with health probes configured at 5s intervals across 20 API pods, this generates ~240 unnecessary metric data points per minute for health checks alone, plus additional noise from Prometheus `/metrics` scrapes.

prein:
**Description**
Pass `excluded_urls` derived from `FILTERED_PATHS` to `FastAPIInstrumentor.instrument_app()`

Resolves #533

### DIFF
--- a/docling_serve/otel_instrumentation.py
+++ b/docling_serve/otel_instrumentation.py
@@ -115,9 +115,16 @@ def setup_otel_instrumentation(
             )
             metrics.set_meter_provider(meter_provider)
 
-    # Instrument FastAPI
-    logger.info("Instrumenting FastAPI with OpenTelemetry")
-    FastAPIInstrumentor.instrument_app(app)
+    # Instrument FastAPI - use excluded_urls to suppress both traces AND metrics
+    # for health/readiness/metrics endpoints. The custom sampler above only filters
+    # trace spans
+    # Patterns are anchored with $ to avoid matching substrings like /healthcheck.
+    excluded_urls = ",".join(f"{p}$" for p in FILTERED_PATHS)
+    logger.info(
+        "Instrumenting FastAPI with OpenTelemetry (excluded_urls=%s)",
+        excluded_urls,
+    )
+    FastAPIInstrumentor.instrument_app(app, excluded_urls=excluded_urls)
 
     # Register RQ metrics if Redis URL is provided
     if redis_url and enable_prometheus:
--- a/tests/test_otel_filtering.py
+++ b/tests/test_otel_filtering.py
@@ -1,0 +1,203 @@
+"""Tests for OpenTelemetry trace filtering of health and metrics endpoints."""
+
+from unittest.mock import MagicMock, patch
+
+import pytest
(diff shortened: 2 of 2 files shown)


dolfim-ibm (approved):
lgtm

PR #538, opened 2026-03-06
feat: ready endpoint for API pods

### LINKED ISSUE #537: Add readiness endpoint to API pod
### Description

API pod has no readiness probe endpoint. Kubernetes has no way to distinguish "app is listening" from "app is actually able to serve requests."
In LocalOrchestrator deployments, the pod accepts traffic before `warm_up_caches()` finishes loading ML models. In RQOrchestrator deployments, the pod accepts traffic before it can reach Redis to enqueue jobs. Both scenarios cause user-visible errors during rollouts.

#### 1. `/ready` endpoint

Add a `GET /ready` endpoint that returns 200 when the pod is genuinely ready to serve:

- **LocalOrchestrator:** gates on `warm_up_caches()` completion (ML model loading)
- **RQOrchestrator:** gates on Redis connectivity (`PING` against the connection pool)
- Returns 503 with a descriptive `detail` message when not ready

prein:
**Description**

Adds a `GET /ready` endpoint that returns 200 when the pod is genuinely ready to serve:

- **LocalOrchestrator:** gates on `warm_up_caches()` completion (ML model loading)
- **RQOrchestrator:** gates on Redis connectivity (`PING` against the connection pool)
- Returns 503 with a descriptive `detail` message when not ready

Also adds `/readyz` (alias) and `/livez` (lightweight liveness, equivalent to `/health`) for Kubernetes convention compatibility.

The existing `load_models_at_boot` setting already controls whether `warm_up_caches()` runs. The readiness gate respects this - if models are loaded at boot, the pod isn't ready until they finish; if not, the gate passes immediately.

**Architecture note:** In RQ deployments, `warm_up_caches()` is a no-op on the API pod (models live in worker pods). The readiness gate fires instantly and only the Redis check is meaningful. The model-loading gate exists for LocalOrchestrator users who run conversion in-process.

Updated the OpenShift deploy example to use:
- `startupProbe` → `/ready` (allows slow model loading without killing the pod)
- `readinessProbe` → `/ready` (gates traffic on actual readiness)
- `livenessProbe` → `/health` (lightweight, doesn't check dependencies)

**Impact**

| Deployment type | Benefit |
|-----------------|---------|
| LocalOrchestrator | Prevents traffic before models load - eliminates timeout errors during rollouts |
| RQOrchestrator | Prevents traffic before Redis is reachable - eliminates connection errors during rollouts |

Resolves #537

### DIFF
--- a/docling_serve/app.py
+++ b/docling_serve/app.py
@@ -13,6 +13,7 @@ from io import BytesIO
 from typing import Annotated
 
 import psutil
+import redis.asyncio
 from fastapi import (
     BackgroundTasks,
     Depends,
@@ -78,6 +79,7 @@ from docling_serve.datamodel.responses import (
     HealthCheckResponse,
     MessageKind,
     PresignedUrlConvertDocumentResponse,
+    ReadinessResponse,
     TaskStatusResponse,
     WebsocketMessage,
 )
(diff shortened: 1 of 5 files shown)


prein:
If I read the log right, this failing build workflow just needs to be re-run.

Issue #543, opened 2026-03-22
Use separate port for serving metrics


ultratin:
As is best practice, `/metrics` should be served from another port.

PR #544, opened 2026-03-22
feat: serve metrics on separate port

ultratin:
Resolves https://github.com/docling-project/docling-serve/issues/543
## Summary
  - Add `DOCLING_SERVE_METRICS_PORT` setting to optionally serve `/metrics` on a dedicated HTTP port, separate from the main API server
  - Wraps `start_http_server` with error handling that raises `RuntimeError` if the port is unavailable
  - Adds tests for server startup, no-op when port is `None`, and port-bind failure

## Motivation
Allows infrastructure teams to expose Prometheus metrics on a different port than the API, which is common in Kubernetes deployments where the metrics port is scraped separately and may have different network policies.


ultratin:
@dolfim-ibm apologies, could you approve again? I fixed a linter issue.

PR #546, opened 2026-03-23
feat: add config options for presets and allow yaml config file

dolfim-ibm:


cau-git (approved):

jspast:
Nice! The configuration file examples at `examples/config.yaml` and `examples/config.json` are missing, though...

docling_serve/settings.py
```diff
@@ -37,6 +43,45 @@ class AsyncEngine(str, enum.Enum):
...
+        try:
+            with open(config_path) as f:
+                if config_path.suffix in [".yaml", ".yml"]:
+                    data = yaml.safe_load(f)
+                elif config_path.suffix == ".json":
+                    data = json.load(f)
+                else:
+                    return {}
+            return data if isinstance(data, dict) else {}
+        except Exception:
```
SamD, line 78: is there a way that this can propagate on error,
any small error goes into the void which is real nasty if attempting to find out why your config settings aren't taking hold, or perhaps an endpoint that can dump them

PR #548, opened 2026-03-24
fix: Upgrade docling pining to 2.82.0

cau-git:


dolfim-ibm:
Done in https://github.com/docling-project/docling-serve/pull/553.

PR #583, opened 2026-04-20
feat: Add `/v1/capacity` endpoint, 429 handler, and optional admission gate

### LINKED ISSUE #581: Expose queue capacity endpoint and handle queue-limit rejection
### 1. No capacity visibility for API callers

When using docling-serve behind a service that splits large documents into chunks (e.g. 10-page PDF → 10 conversion jobs), the caller has no way to know whether the system can absorb more work. The only signal today is `task_position` on individual jobs - there is no global capacity view.

The raw data exists internally (RQ: `get_queue_jobs()` / `get_workers_stats()`; Ray: `RedisStateManager` per-tenant counts) and is already scraped by Prometheus, but it is **not accessible to API callers** without a metrics scraping setup.

### 2. `QueueLimitExceededError` returns HTTP 500 (bug)

The Ray orchestrator raises `QueueLimitExceededError` when `enable_queue_limit_rejection=True` and a tenant's queue limit is exceeded.
However, `app.py` does not have an exception handler for this error. It propagates as an **unhandled 500 Internal Server Error** instead of the expected **429 Too Many Requests**.
Note: we're not using Ray, I noticed this was added recently and it looked related.

### A. `GET /v1/capacity` endpoint

Add a lightweight endpoint that returns a system-level capacity snapshot:

```json
{
  "queue_depth": 27,
  "active_jobs": 4,
  "active_workers": 8,
  "max_queue_size": null
}
```

| Field | Description |
|-------|-------------|
| `queue_depth` | Pending jobs not yet started |
| `active_jobs` | Jobs currently being processed |
| `active_workers` | Registered workers (RQ) or actors (Ray) |
| `max_queue_size` | Configured queue limit, if any (`null` = unlimited) |

This requires an abstract `get_capacity()` method on `BaseOrchestrator` in `docling-jobkit`, with concrete implementations for each engine:

- **RQ:** `rq_queue.count`, `started_job_registry.count`, `Worker.all()`
- **Ray:** aggregated `get_tenant_queue_size()`, running task counts, actor count
- **Local:** internal queue size, active threads, `num_workers`
- **KFP:** returns `None` (not applicable)

The snapshot should be **cached with a short TTL** (default 2s) because `Worker.all()` (RQ) and `scan_iter` (Ray) perform Redis scans that are too expensive to run on every request.

### B. `QueueLimitExceededError` → HTTP 429

Add an exception handler in `app.py`:

```python
@app.exception_handler(QueueLimitExceededError)
async def queue_limit_handler(request, exc):
    return JSONResponse(
        status_code=429,
        content={"detail": str(exc)},
        headers={"Retry-After": "5"},
    )
```

This mirrors the existing `RedisBackpressureError` → 503 handler but uses 429 to distinguish queue-level admission rejection from connection-pool saturation.

### C. Optional: RQ queue-level admission gate

The Ray orchestrator has `check_tenant_can_enqueue()` but RQ has no queue-level admission control. If `queue_max_size` is configured, check `orchestrator.queue_size()` before `enqueue()` and reject with 429 when the queue is full. This activates the existing `queue_max_size` setting as an admission gate (it currently only controls internal pipeline batching).

## Use Case

Our workflow:

1. User uploads a PDF to the extraction service.
2. Extraction service splits it into chunks (one per page/section).
3. Each chunk is submitted to docling-serve as an async conversion job.
4. Extraction service polls for each job's status.

Without a capacity signal, step 3 submits all chunks as fast as possible.
With `/v1/capacity`, the extraction service can:

- Check `queue_depth / active_workers` before each batch submission
- Pause when `active_workers == 0` (no workers available)
- Throttle when the ratio exceeds a client-side threshold
- Resume once the system drains

This is application-level back pressure that works within the existing polling architecture, without infrastructure changes.

## Happy to Contribute

I'm prepared to submit PRs for both `docling-serve` and `docling-jobkit` if the approach looks reasonable.

prein:
Depends on docling-project/docling-jobkit#126 (`SystemCapacity` + `get_capacity()`).

See #581  for full problem descriptio

## Summary

- **`GET /v1/capacity`** - returns system-level capacity snapshot (queue depth, active jobs, worker count); cached with configurable TTL (default 2s) to avoid per-request Redis scans
- **`QueueLimitExceededError` → 429** - fixes unhandled 500 when Ray queue limit rejection is enabled; returns capacity snapshot in response body
- **Optional admission gate** - new `admission_max_queue_size` setting; when set, rejects submissions with 429 before `enqueue()` if queue is full

Resolves #581

### DIFF
--- a/docling_serve/app.py
+++ b/docling_serve/app.py
@@ -10,7 +10,7 @@ import time
 from collections import Counter
 from contextlib import asynccontextmanager
 from io import BytesIO
-from typing import Annotated
+from typing import Annotated, Optional
 
 import psutil
 from fastapi import (
@@ -84,11 +84,14 @@ from docling_jobkit.orchestrators.base_orchestrator import (
     BaseOrchestrator,
     ProgressInvalid,
     RedisBackpressureError,
+    SystemCapacity,
     TaskNotFoundError,
 )
+from docling_jobkit.orchestrators.ray.orchestrator import QueueLimitExceededError
 from docling_jobkit.orchestrators.rq.orchestrator import RQOrchestrator
 
 from docling_serve.auth import APIKeyAuth, AuthenticationResult
+from docling_serve.capacity import get_cached_capacity
 from docling_serve.helper_functions import DOCLING_VERSIONS, FormDepends
 from docling_serve.orchestrator_factory import get_async_orchestrator
 from docling_serve.otel_instrumentation import (
@@ -237,6 +240,21 @@ def create_app():  # noqa: C901
             status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
             content={"detail": "Server is busy, please try again shortly."},
             headers={"Retry-After": "1"},
+        )
+
+    @app.exception_handler(QueueLimitExceededError)
+    async def queue_limit_exceeded_handler(
+        request: Request, exc: QueueLimitExceededError
+    ) -> JSONResponse:
+        del request
+        capacity = await get_cached_capacity(get_async_orchestrator())
+        return JSONResponse(
+            status_code=status.HTTP_429_TOO_MANY_REQUESTS,
+            content={
+                "detail": str(exc),
+                "capacity": capacity.model_dump() if capacity else None,
+            },
+            headers={"Retry-After": "5"},
         )
 
     # Setup OpenTelemetry instrumentation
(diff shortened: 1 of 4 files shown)


prein:
We are running it in our preprod environment and it works as expected. It would be amazing if you could find time to review it and hopefully merge and release so we don't have to maintain it in a fork.

k-sparrow:
any news on this?

PR #605, opened 2026-05-12
feat: Control error detail of public API responses 

cau-git:
## Summary
This change hardens `docling-serve`'s public error surfaces so infrastructure exception details are no longer exposed by default.

A new deployment flag, `DOCLING_SERVE_DEBUG_ERROR_DETAILS`, controls whether raw exception text is returned:
- `false` (default): public HTTP error details are sanitized
- `true`: current raw exception detail is preserved for debugging

## Why
Several service-layer error paths were returning `str(exc)` directly in public responses. That can leak backend/runtime internals such as Ray exception text, wrapper errors, and operational details that are useful in logs but should not be part of the default API surface.

This change keeps raw detail available in debug mode while making the default production behavior safer and more stable.

## What changed
- Added `debug_error_details: bool = False` to `DoclingServeSettings`
- Documented `DOCLING_SERVE_DEBUG_ERROR_DETAILS` in `docs/configuration.md`
- Added a small helper for rendering public HTTP error details consistently
- Sanitized the main audited HTTP error paths:
  - readiness failures
  - internal progress callback validation failures
  - Ray `DispatcherUnavailableError` 503 responses
- Threaded the debug flag into Ray and RQ orchestrator config creation so downstream jobkit sanitization can use the same deployment control

## Behavior
With debug disabled:
- readiness returns a generic `Readiness check failed`
- invalid internal progress callback payloads return `Invalid progress payload.`
- Ray dispatcher-unavailable responses return `Ray dispatcher is unavailable.`

With debug enabled:
- those same responses surface the underlying exception detail again

## Scope
This PR changes error value rendering only. It does not change response schemas or endpoint contracts.

### DIFF
--- a/docling_serve/app.py
+++ b/docling_serve/app.py
@@ -104,6 +104,7 @@ from docling_serve.policy import (
     validate_convert_options,
     validate_convert_request,
 )
+from docling_serve.public_errors import build_public_http_detail
 from docling_serve.response_preparation import prepare_response
 from docling_serve.settings import AsyncEngine, docling_serve_settings
 from docling_serve.storage import get_scratch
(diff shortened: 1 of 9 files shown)


cau-git:
superseded by #609

PR #610, opened 2026-05-20
feat: RQ worker hooks and tracing support

cau-git:
## Summary

- Wire the new jobkit hook API into `instrumented_docling_task` to deliver
  per-phase OTel sub-spans across the full RQ job lifecycle: source preparation
  (with per-file md5/size span events), document conversion, result processing,
  serialization, and cleanup
- Replace a previous inline implementation that duplicated jobkit internals with
  a clean callback-based approach: execution logic stays in jobkit, instrumentation
  logic stays in docling-serve

see https://github.com/docling-project/docling-jobkit/pull/152


cau-git:
obsolete, closing.

PR #618, opened 2026-05-30
feat: add support for ROCm 7.2

### LINKED ISSUE #475: Support for AMD ROCm 7.10 (strix halo)
Hello,
since AMD Strix Halo is becoming more popular, can you add docker support for rocm7.10?
I can contribute as well

kresimirfijacko:
Adds a new `rocm72` variant of the docling-serve container image targeting AMD GPUs on ROCm 7.2 (tested on AMD Strix Halo iGPU).

## Changes
1. Added rocm7.2 to pyproject.toml, Makefile and compose-amd.yaml
2. added `libatomic` to os-packages.txt because it is required by PyTorch/ROCm runtime (AMD wheels do not bundle it?)

## Notes:
I am not very experienced with docling, i tried couple of arxiv pdf's and they worked. Docling provides several modes:
1. standard pipeline - work ok, but triggers RapidOCR. By default it uses onnxruntime. There is onxxruntime-migraphx targeting AMD ROCm but atm RapidOCR doesn't support it out of the box. I tried to use for sake of experiment in different environment and it didn't provide any benefit (i guess models are too small?)
2. VLM pipeline, using `granite-docling-258M` i had stability/performance issues, and not quite sure what is problem. I tried to run it in separate environment but had no success. I think it's better to point to remote VLM model (using some other inference engine, in strixhalo it would be llama.cpp).

Resolves #475


dolfim-ibm (approved):
lgtm

PR #629, opened 2026-06-11
feat: Add settings for coordinator and converter max_replicas_per_node

cau-git:
## Summary

Surfaces the new per-node replica cap from docling-jobkit as docling-serve settings, so operators can control coordinator/converter placement across Ray nodes.

## Changes

- **`docling_serve/settings.py`.** Added `eng_ray_converter_max_replicas_per_node` and `eng_ray_coordinator_max_replicas_per_node` (`Optional[int]`, default `None`).
- **`docling_serve/orchestrator_factory.py`.** Mapped both settings onto the corresponding `RayOrchestratorConfig` fields (`converter_max_replicas_per_node`, `coordinator_max_replicas_per_node`).

## Behavior

Default is `None` for both, i.e. no cap and no change to existing deployments. Setting `eng_ray_coordinator_max_replicas_per_node=1` spreads coordinators one per Ray node; the converter cap is typically left unset so Ray packs converters for density.

## Dependency
see https://github.com/docling-project/docling-jobkit/pull/173


PeterStaar-IBM (approved):
lgtm!

PR #630, opened 2026-06-12
fix: supervise orchestrator queue processor and harden WebSocket status push

cau-git:
PR summary

  ## Problem

  When the orchestrator's background pub/sub listener died, the service kept reporting healthy indefinitely (`All connections healthy` logged hundreds of times) while
  WebSocket status push was silently broken — clients hung on `started` and only HTTP polling worked. Nothing supervised the long-running `process_queue()` task, and
  the WebSocket request/response loop held a stale task snapshot that never closed on completion, so a missed push left the client stuck forever.
  
  ## Changes

  - **Supervise the background queue processor.** `process_queue()` is expected to run for the lifetime of the app; it is now tracked with a done-callback that, if the
  task ever finishes or raises (excluding shutdown cancellation), logs the failure with a traceback and sets a `_queue_processor_failed` flag.
  - **Reflect orchestrator-loop health in the probes.** `/livez` now returns 503 when the queue processor has died so the platform restarts the pod (which
  re-subscribes the pub/sub listener), and `/ready` (and `/readyz`) gate on the same flag to pull the pod out of rotation. Previously a dead listener was invisible to
  both probes.
  - **Harden the WebSocket loop (defense-in-depth).** The status loop now re-fetches the task each iteration instead of reusing the snapshot taken at connect time, and
  closes the socket once `task.is_completed()`. So even if a real-time push is missed, the client's next poll sees current state and the connection terminates cleanly
  instead of streaming `started` forever. A `TaskNotFoundError` raised mid-stream (e.g. a reaped task) now closes the socket gracefully.

### DIFF
--- a/docling_serve/app.py
+++ b/docling_serve/app.py
@@ -147,6 +147,31 @@ _log = logging.getLogger(__name__)
 # implementation is a no-op so this event fires instantly in RQ deployments.
 _models_ready = asyncio.Event()
 
+# Set if the background queue processor task dies with an error. Liveness/
+# readiness then fail so the platform restarts the pod instead of silently
+# serving with a dead orchestrator loop: a dead pub/sub listener stops WebSocket
+# push delivery while polling still succeeds, which is otherwise very hard to
+# detect.
+_queue_processor_failed = asyncio.Event()
+
+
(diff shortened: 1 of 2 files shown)


dolfim-ibm (approved):
lgtm

PR #634, opened 2026-06-14
fix: construct s3_presigned_config in RQ worker entrypoint

### LINKED ISSUE #633: `PresignedUrlTarget` fails in RQ engine: worker entrypoint doesn't build s3_presigned_config
> PresignedUrlTarget requires s3_presigned_config in orchestrator config

  The API admits the request (policy passes when `artifact_storage_enabled=true`), enqueues it, and the worker rejects it at the point of artifact upload.

  ## Root cause

  API and worker each construct their own `RQOrchestratorConfig` in separate processes.

  - API path (`docling_serve/orchestrator_factory.py:127-152`): calls `_build_s3_presigned_config()` and passes the result as `s3_presigned_config=...`.
  - Worker path (`docling_serve/__main__.py:rq_worker()`): constructs `RQOrchestratorConfig` inline at lines 423-437 and omits the field. The field defaults to `None`.

  When the task reaches `S3PresignedTargetProcessor` in the worker, jobkit's check raises `PresignedUrlTarget requires s3_presigned_config in orchestrator config`. The worker exception handler then publishes
  `FAILURE` without an error_message, so the client sees the generic `Internal processing error.` (jobkit's `INTERNAL_TASK_ERROR_MESSAGE`).

  ## Reproduction

  1. Deploy v1.22.x with the RQ engine (separate API and worker pods).
  2. Set on **both** pods:
     - `DOCLING_SERVE_ARTIFACT_STORAGE_ENABLED=true`
     - `DOCLING_SERVE_ARTIFACT_STORAGE_BUCKET=<bucket>`
     - `DOCLING_SERVE_ARTIFACT_STORAGE_ACCESS_KEY=<...>`
     - `DOCLING_SERVE_ARTIFACT_STORAGE_SECRET_KEY=<...>`
  3. Submit a convert request with `{"target": {"kind": "presigned_url"}}`.
  4. Status polls return SUCCESS instantly with all-zero `task_meta`; result polls show `task_status=failure` with `error_message="Internal processing error."`.
  5. Worker logs show: `PresignedUrlTarget requires s3_presigned_config in orchestrator config`.

  ## Affected engines

  - **RQ**: broken (this issue).
  - **Local**: unaffected — single process, `orchestrator_factory.py:55` already populates the field on `LocalOrchestratorConfig`.
  - **Ray**: unaffected — Ray's worker dispatch inherits config through actor serialisation, not via a separate CLI entrypoint that builds its own.

  ## Suggested fix

  Have the worker entrypoint call the same `_build_s3_presigned_config()` helper the API uses, and pass the result into `RQOrchestratorConfig`. ~3 lines.

  I have a working patch and have verified the fix end-to-end against my own test environment — happy to send a PR.

dan-m62:
  Resolves #633

v1.22 added `PresignedUrlTarget` but the RQ worker `entrypoint in __main__.rq_worker()` never builds the operator-managed `S3PresignedConfig`. 

The API process builds it (`orchestrator_factory.get_async_orchestrator calls _build_s3_presigned_config()` and passes it into `RQOrchestratorConfig` at three call sites), but the worker process constructs its own `RQOrchestratorConfig` inline and omits the field, so it defaults to `None`.

Result: any task with `target.kind=presigned_url` is accepted by the API (policy admission passes because `artifact_storage_enabled=true`), enqueued into RQ, then fails in the worker with:

    PresignedUrlTarget requires s3_presigned_config in orchestrator config

The error message surfaces to the client as the generic "Internal processing error." (jobkit's `INTERNAL_TASK_ERROR_MESSAGE`) since the worker exception handler publishes `FAILURE` without an error_message.

The `S3PresignedConfig` is operator policy, intentionally not transmitted in task payloads (would put AWS credentials into queue messages). Each process is meant to build it independently from its own environment — which is what the API path already does. This commit applies the same construction to the worker entrypoint by importing the existing `_build_s3_presigned_config()` helper.

Verified by reproducing the original failure against a private test cluster (docling-serve v1.22.0, RQ engine, separate API and worker pods, DOCLING_SERVE_ARTIFACT_STORAGE_* env vars set on both), then applying this patch as an overlay image and re-running the same convert request. After the fix the worker accepts the task, uploads artifacts to the configured bucket, and the API returns presigned `GET` URLs which serve the converted documents.

This affects only the RQ engine. The local engine builds the config correctly via `orchestrator_factory.py:55`, and Ray's worker dispatch inherits the config through Ray's actor serialisation.

### DIFF
--- a/docling_serve/__main__.py
+++ b/docling_serve/__main__.py
@@ -394,6 +394,7 @@ def rq_worker() -> Any:
     )
 
     from docling_serve.logging_config import setup_logging
+    from docling_serve.orchestrator_factory import _build_s3_presigned_config
     from docling_serve.rq_instrumentation import setup_rq_worker_instrumentation
     from docling_serve.rq_worker_instrumented import InstrumentedRQWorker
 
@@ -427,6 +428,7 @@ def rq_worker() -> Any:
         redis_gate_reserved_connections=docling_serve_settings.eng_rq_redis_gate_reserved_connections,
         redis_gate_wait_timeout=docling_serve_settings.eng_rq_redis_gate_wait_timeout,
         redis_gate_status_poll_wait_timeout=docling_serve_settings.eng_rq_redis_gate_status_poll_wait_timeout,
+        s3_presigned_config=_build_s3_presigned_config(),
     )
 
     cm_config = DoclingConverterManagerConfig(


dolfim-ibm (approved):
lgtm

cau-git (approved):

PR #635, opened 2026-06-15
fix: resolve omitted convert target from server policy

cau-git:
## Problem

`ConvertSourcesRequest.target` declares a model-level default (`PresignedUrlTarget` in the unreleased docling, `InBodyTarget` before that). Clients drop fields left at their default, so when `target` is omitted the request arrives carrying whatever kind the model declares — independent of what the deployment actually allows. A non-presigned deployment then 422-rejects an omitted target, and a presigned-only deployment that disallows inbody would reject the other way. The default was a single static value baked into a shared type, so no single choice is correct across deployment topologies.

## Fix

`resolve_default_target(policy)` in `policy.py` derives the omitted-target default from server policy: it prefers `presigned_url` whenever artifact storage is enabled (keeping result payloads out of the response body) and otherwise falls back to `inbody`. The selection and validation both live in `policy.py`, so policy remains the single source of truth.

The default is injected where the policy is known, inside `create_app()`:
- JSON body endpoints (`/v1/convert/source`, `/v1/convert/source/async`) bind a `create_model` subclass of `ConvertSourcesRequest` that repopulates the `target` default with the resolved value. The subclass preserves the public `ConvertSourcesRequest` schema name and remains an instance of the base, so downstream `isinstance` checks and the jobkit handoff are unaffected.
- Multipart file endpoints (`/v1/convert/file`, `/v1/convert/file/async`) use the resolved `TargetName` as the `target_type` form default.

Because the resolved default is a concrete value, it renders directly into the OpenAPI schema (both the body `target.default` and the form `target_type.default`) per deployment — no spec post-processing or downgrade-handler changes.

## Behavior

- Omitted target → server-resolved default (always policy-valid), and the injected default is still re-validated by `_prepare_convert_request`.
- Explicitly specified target → passed through unchanged and validated; respected if allowed, 422-rejected if not. It is never silently coerced to the default.
- Batch endpoint unchanged (`target` is required, omission still 422s as a missing field).
- Chunk endpoints unchanged (presigned is rejected there; `inbody` stays the only sensible default).

## What crosses into jobkit

Only `request.target` — a plain docling target instance and a member of jobkit's `TaskTarget` union — is enqueued. The request-model subclass stays in the serve layer; jobkit receives the same `kind`-discriminated target it always did.


dolfim-ibm (approved):
lgtm

PR #647, opened 2026-07-04
feat: add include_provenance option to chunk endpoints

### LINKED ISSUE #613: Include provenance (bbox + page_no) inline in chunk endpoint response
The /v1/chunk/hybrid/file and /v1/chunk/hybrid/source endpoints currently return doc_items as JSON Pointer references (e.g. "#/texts/0") without resolving the underlying provenance data. As a result, bounding box and page number information is not available in the chunk response.

Page numbers are surfaced separately as a flat page_numbers array, but bounding boxes are completely absent. The only workaround is to make a second call to the convert endpoint with to_formats: ["json"] and resolve the refs client-side — meaning the document goes through the full AI processing pipeline twice, which doubles latency.

This is particularly painful because docling's conversion is the slow part. The chunking itself is fast. The server already has the fully resolved DoclingDocument in memory when it builds the chunk response — it just doesn't include the prov data in the output.

**Requested change**
When serializing doc_items in the chunk response, include the resolved prov array for each item inline, rather than returning bare refs. Concretely, instead of: 

`"doc_items": ["#/texts/0"] `

return:

```
"doc_items": [
  {
    "self_ref": "#/texts/0",
    "label": "text",
    "prov": [
      {
        "page_no": 1,
        "bbox": { "l": 108.0, "t": 405.1, "r": 504.0, "b": 330.7, "coord_origin": "BOTTOMLEFT" },
        "charspan": [0, 608]
      }
    ]
  }
]
```

Alternatively, a flag like chunking_include_provenance: true would be acceptable if there's concern about response size, though given that chunks already carry text content, the bbox data is small by comparison.

**Use case**
RAG pipelines that need to highlight or cite source regions in the original document (e.g. rendering a bounding box overlay on the PDF page) currently cannot do so without a double-pass. This is a common requirement for document grounding in RAG applications.

Thank you for considering, this extension would be really heplfull.

pablopupo:
Chunk responses currently expose `doc_items` as JSON pointer references, so clients cannot obtain bounding boxes without converting the document again and resolving the references themselves.

This adds an opt-in `include_provenance` chunking option (`chunking_include_provenance` for multipart requests). When enabled, each `doc_items` entry contains its `self_ref`, `label`, and resolved `prov` array. The converted document is used internally to resolve those references and omitted from the public response unless `include_converted_doc` was also requested.

The option applies to hybrid and hierarchical chunking across synchronous and asynchronous file and source routes. Requests using ZIP output, referenced image exports, or the Ray engine return 422 before enqueue. Ray recovery does not currently retain the response metadata needed to distinguish an internal document from one requested by the caller.

The flag remains off by default, and the existing response shape is unchanged.

Resolves #613

## Verification

- 61 provenance resolver, endpoint, stored-result, and OpenAPI tests
- Ruff formatting and lint checks
- MyPy across `docling_serve`
- `uv lock --check`
- Full pre-commit suite
- Source distribution and wheel builds

### DIFF
--- a/docling_serve/app.py
+++ b/docling_serve/app.py
@@ -10,7 +10,7 @@ import time
 from collections import Counter
 from contextlib import asynccontextmanager
 from io import BytesIO
-from typing import Annotated, Any
+from typing import Annotated, Any, cast
 
 import psutil
 from fastapi import (
@@ -47,8 +47,6 @@ from docling.datamodel.service.callbacks import (
 )
 from docling.datamodel.service.chunking import (
     BaseChunkerOptions,
-    HierarchicalChunkerOptions,
-    HybridChunkerOptions,
 )
 from docling.datamodel.service.options import (
     ConvertDocumentsOptions as ConvertDocumentsRequestOptions,
@@ -62,9 +60,9 @@ from docling.datamodel.service.requests import (
     make_request_model,
 )
 from docling.datamodel.service.responses import (
-    ChunkDocumentResponse,
     ClearResponse,
     ConvertDocumentResponse,
+    DoclingTaskResult,
     HealthCheckResponse,
     MessageKind,
     PresignedUrlConvertDocumentResponse,
@@ -80,6 +78,7 @@ from docling.datamodel.service.targets import (
     ZipTarget,
 )
 from docling.datamodel.service.tasks import TaskType
+from docling_core.types.doc.document import ImageRefMode
 from docling_jobkit.connectors.errors import (
     SourceConnectorConfigError,
     TargetConnectorConfigError,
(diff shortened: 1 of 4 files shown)


dolfim-ibm:
@pablopupo we recently added the chunks as export modes of the normal convert endpoint. Additionally, also allowing multiple `targets` (e.g. one for the chunks, one for doc exports). With this change we are also soon planning to deprecate the chunking endpoint.

Do you think your proposed feature is still needed?

pablopupo:
Yes, the new convert path addresses the original problem. Requesting chunks and JSON together removes the second conversion pass, and with the chunk endpoints headed for deprecation this PR is no longer needed. Thanks for pointing me to the new flow.

PR #651, opened 2026-07-13
feat: add allowed_source_types policy control

cau-git:
## Add `allowed_source_types` policy control

Deployments could restrict result destinations with `allowed_target_types`, but nothing constrained source kinds — the endpoints accepted anything in the request union types. This adds a symmetric `allowed_source_types` control.

### What changed

- New `allowed_source_types` setting (`Optional[list[str]]`, defaults to all allowed), parsed from CSV/JSON env like the other allow-lists.
- `ALL_SOURCE_TYPES` is derived at import time from the actual endpoint source unions (`SourceRequestItem`, `BatchSourceRequestItem`) rather than hardcoded, so it tracks the request models and a configured list can only intersect it — never extend beyond `{file, http, s3}`.
- `validate_source_kinds` enforces the allow-list on the convert, batch-convert, and chunk endpoints, returning 422 for a disallowed source kind.

### Notes

- Existing S3 source/target pairing checks are unchanged and still run; disallowing `s3` just rejects earlier with a clearer message.
- Default behavior is unchanged: with no setting configured, all source kinds remain allowed.

### DIFF
--- a/docling_serve/policy.py
+++ b/docling_serve/policy.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import typing
 from dataclasses import dataclass
 from typing import Any, TypeVar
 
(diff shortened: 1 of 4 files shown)


dolfim-ibm (approved):
lgtm

PR #652, opened 2026-07-13
feat: Accept GCS, Azure Blob, and Google Drive sources/targets

cau-git:
Docling now carries GCS, Azure Blob, and Google Drive in `BatchSourceRequestItem` and the target unions. The batch/source endpoints passively accept the new source schemas, but the request-to-task dispatch must map them explicitly, so this wires that up.

### Changes
- **app.py** — `_enqueue_source` gains branches mapping `azure_blob`, `google_cloud_storage`, and `google_drive` request items to their coord models (same shape as the existing S3 branch).
- **app.py** — added a terminal `else` guard in the dispatch loop: an unmapped source kind now raises instead of being silently dropped (the previous `if/elif` chain had no fallback).
- **app.py** — renamed `_enque_source` → `_enqueue_source`.
- **policy.py** — `ALL_TARGET_TYPES` now includes `azure_blob`, `google_cloud_storage`, `google_drive`, so `validate_target_kind` admits the new targets.
- **policy.py** — generalize the S3 source→target pairing rule: any expandable source (s3, azure_blob, google_cloud_storage, google_drive) now requires a storage target (same set), via `validate_source_target_pairing`. Drops the old S3-only reverse constraint, so a file/http source may now write to a storage target.
- **pyproject.toml** — `[tool.uv.sources]` pins `docling-slim` to the docling branch below.

### Notes
- Depends on docling branch **`cau/move-connector-types-gcs-gd-azure`** (and the matching docling-jobkit change).
- **Google Drive is batch-only.** It is not in `SourceRequestItem`, so the single-document convert endpoint rejects it (opaque `path_id` can't be validated as file-vs-folder).
- Source→target pairing rules were left unconstrained for the new kinds (only the existing S3 pairing rule remains).


docs/usage.md
```diff
@@ -81,8 +81,9 @@ On top of the source of file (see below), both endpoints support the same parame
 | `api_overrides` | Dict[VlmEngineType, ApiModelConfig] | API-specific configuration overrides |
 | `trust_remote_code` | bool | Whether to trust remote code for this model |
 | `stop_strings` | List[str] | Stop strings for generation |
+| `temperature` | float | Sampling temperature for generation |
 | `max_new_tokens` | int | Maximum number of new tokens to generate |
-| `temperature` | float | Sampling temperature for generation. 0.0 uses greedy decoding. |
+| `extra_generation_config` | Dict[str, Any] | Additional generation configuration |
```
ceberam, line 86: Tiny comment: since we are using the built-in collection types for annotations instead of those from `typing`, I would write `dict[str, Any]` (and update the others in this list).

ceberam (commented):
Just a small comment

docs/usage.md
cau-git: Let's better do it on a separate PR, this documentation has Dict[str, Any] all over the place.

ceberam (approved):

PR #655, opened 2026-07-17
feat: add Azure Blob presigned artifact storage

cau-git:
## Summary

Adds Azure Blob Storage as a backend for server-managed `PresignedUrlTarget` artifacts.

- Adds `s3` and `azure` managed artifact backend selection.
- Adds Azure connection-string, account, container, blob-prefix, and SAS lifetime settings.
- Validates missing Azure configuration using the operator-facing environment-variable names.
- Builds `AzurePresignedConfig` and passes the backend-neutral presigned configuration through Local, RQ, and Ray orchestration, including the direct RQ worker entry point.
- Documents the S3 and Azure managed-storage settings and the account-key-only Azure scope.

## Dependency

Depends on docling-project/docling-jobkit#204. Until that change is released, `uv.lock` pins the Jobkit feature branch at commit `d4717cf687354e3a87784cc5f55696b3b1dc9239`; the Git source should be replaced with the released minimum version before merge.

## Testing

Coverage includes:

- Azure managed-storage settings and configuration validation
- missing-setting and unsupported-backend errors
- direct RQ worker configuration
- Local, RQ, and Ray presigned configuration wiring


dolfim-ibm (approved):
lgtm

PR #662, opened 2026-07-25
feat(UI): expose PDF heading-level inference

### LINKED ISSUE #646: Add options for heading level inference
https://github.com/docling-project/docling/pull/3633 and https://github.com/docling-project/docling/pull/3688 added options for setting the header  levels from the PDF contents, so that we don't just get header level 1 everywhere.
This is by default off and configureable in the PdfPipelineOptions.
Can you please add the necessary options to the Docling Serve API?

DanielNg0729:
Resolves #646

docling-project/docling#3633 and docling-project/docling#3688 added inference of section-header levels from PDF bookmarks, outline numbering and font style, configurable via `PdfPipelineOptions.heading_hierarchy_options` and off by default. Those options were not reachable from the service API, so every heading returned by docling-serve stayed at `level=1` and the document hierarchy came back flat.

The API options model now lives in docling core and the pipeline mapping in docling-jobkit, so exposing this takes three coordinated PRs. Merge order: docling → docling-jobkit → **docling-serve**.

**This PR** adds an "Infer heading levels" toggle to the Gradio UI and regenerates `docs/usage.md` with the new parameter and the nested `HeadingHierarchyOptions` table.

It also fixes `_format_type` in `scripts/update_doc_usage.py`: PEP 604 unions (`X | None`) carry no `__origin__`, so they fell through to `str()` and the literal pipe split the surrounding markdown table row. This is not cosmetic — it breaks the rendered table, and it would have corrupted the new `numbering_schemes` row plus three pre-existing rows on the next regeneration.

**Usage**

```json
{
  "options": {
    "to_formats": ["json"],
    "heading_hierarchy_options": { "enabled": true }
  }
}
```

Advanced knobs (`use_bookmarks`, `use_numbering`, `use_style`, `numbering_schemes`, `max_level`, `bookmark_match_threshold`) are documented in `docs/usage.md`. The UI exposes only the on/off toggle, consistent with how it surfaces other options.

**Verification**

`tests/test_heading_hierarchy_options.py` covers the boundary this repo owns: that the nested option survives the `multipart/form-data` path. Form fields cannot carry objects, so `FormDepends` accepts nested models as JSON strings — an option never exercised through that path can silently degrade to its default while the JSON-body path keeps working.

End-to-end, converting `2408.09869v5.pdf` through `/v1/convert/file`:

| | section-header levels |
|---|---|
| option omitted (default) | `{1: 20}` |
| `{"enabled": true}` | `{1: 8, 2: 6, 3: 4, 4: 2}` |

`3.1 PDF backends` → level 2, `Layout Analysis Model` → level 3, `Abstract` → level 2. 12 of 20 headings corrected, and the default path is unchanged.

**Note on a pre-existing bug found while testing**

`FormDepends` wraps validation failures in a plain `ValueError`, which FastAPI does not translate, so malformed nested options on `/v1/convert/file` return **500 instead of 422**. This affects `picture_description_local`, `vlm_pipeline_model_*` and any other nested field, not just the one added here. Left out of this PR as an unrelated fix; happy to open a separate issue.


dolfim-ibm (approved):
lgtm

PR #665, opened 2026-07-28
fix: support target not just target type in multipart requests

phact:
The multipart file conversion endpoints only take target_type, rather than the full target object. This works fine for some response targets but breaks for storage targets that need to pass metadata: 

```sh
curl -X POST 'http://127.0.0.1:5001/v1/convert/file/async' \
  -F 'files=@document.pdf;type=application/pdf' \
  -F 'target_type=s3' \
  -F 'target={"kind":"s3","endpoint":"s3.example.com","access_key":"ACCESS_KEY","secret_key":"SECRET_KEY","bucket":"converted","key_prefix":"results/"}'
```

For compatibility, `target_type` can still be used by itself for the `inbody`, `zip`, and `presigned_url` targets.

**Checklist:**

- [x] Documentation has been updated, if necessary.
- [x] Examples have been added, if necessary.
- [x] Tests have been added, if necessary.

### DIFF
--- a/docling_serve/app.py
+++ b/docling_serve/app.py
@@ -36,7 +36,7 @@ from fastapi.openapi.docs import (
 )
 from fastapi.responses import JSONResponse, RedirectResponse
 from fastapi.staticfiles import StaticFiles
-from pydantic import create_model
+from pydantic import TypeAdapter, ValidationError, create_model
 from scalar_fastapi import get_scalar_api_reference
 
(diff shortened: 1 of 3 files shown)


vku-ibm:
@phact please resolve DCO check, otherwise looks good

dolfim-ibm:
closing as discussed in https://github.com/docling-project/docling/pull/3907

PR #670, opened 2026-08-06
fix: new transformers for linux

dolfim-ibm:
Docling has a conditional dependency for `transformers` because this [issue](https://github.com/huggingface/transformers/issues/46159) causes malfunctions on macOS.

Using some lock constrains we force uv to install newer versions when possible, for example on Linux, i.e. in the compiled images.


vku-ibm (approved):

kdinkla (approved):
lgtm

PR #680, opened 2026-08-18
fix(form): keep model_fields_set truthful on the multipart endpoints (#674)

### LINKED ISSUE #674: `ocr_engine` accepts unvalidated garbage values and silently ignores them (still present in 1.29)
This is the underlying defect behind #567 and #554, which were closed on a workaround (`use ocr_preset instead`) rather than a fix. The workaround is correct and we use it. This report is about the part that was never addressed: **`ocr_engine` accepts values that do not exist and returns 200.**

A maintainer could not reproduce the original reports. The reproducer below should settle it, because it does not depend on which engine ends up running — it only asks whether the API validates its input at all.

## Reproducer

docling-serve **1.29.0**, `/v1/convert/file/async`, same 3-page scanned PDF, one field varied:

| field sent | HTTP | chars returned | honoured |
|---|---:|---:|---|
| `ocr_engine=tesseract` | 200 | 3194 | no |
| `ocr_engine=nonsense_engine` | 200 | 3194 | **no — nonexistent engine accepted** |
| `ocr_preset=tesseract` | 200 | 3160 | yes |
| `ocr_preset=easyocr` | 200 | 2579 | yes |
| `ocr_preset=nonsense_preset` | **422** | — | yes, properly validated |

`ocr_preset` validates and rejects garbage. `ocr_engine` accepts a value that cannot possibly be an engine, returns 200, and produces output byte-identical to sending nothing at all. It is typed as a bare `str`, marked deprecated in the docs, and then ignored by the pipeline.

## Why this is worth fixing rather than documenting

Silent acceptance is what makes it expensive. A deprecated-but-honoured field is fine; a deprecated-and-ignored field that still returns 200 costs real work downstream:

- We ran a full OCR engine comparison sweeping `ocr_engine` and concluded "engine choice does not affect speed". Every arm had run the same engine. Our collaborators independently reached the same wrong conclusion from the same cause.
- Once `ocr_preset` is used, the engines differ by **2.5x** (2.33 vs 5.92 s/page). That difference had been invisible to us for weeks.
- The identical character counts across arms are the only tell, and you only notice them if you happen to compare output sizes.

**The server log actively hides it.** docling emits `auto_ocr_model - Auto OCR model selected rapidocr` while warming its *default* converter, which is indistinguishable from honouring a request. Output content is the only reliable witness.

## Suggested fix

Validate `ocr_engine` against the same enum as `ocr_preset`, then either honour it or return 422. Either is fine. Accepting `nonsense_engine` with a 200 is the harmful behaviour, and it is a small change relative to how much confusion the field has generated across #554, #567 and this report.

Anai-Guo:
Fixes #674.

### What is wrong

On the multipart endpoints (`/v1/convert/file`, `/v1/convert/file/async`, the chunk endpoints), `ocr_engine` is accepted with any value, returns 200, and is then ignored — output is byte-identical to sending nothing. `ocr_preset` on the same request rejects an unknown value with 422 and is honoured.

The two options are supposed to agree: `ConvertDocumentsOptions.validate_ocr_options` forwards the deprecated `ocr_engine` onto `ocr_preset` when the caller set `ocr_engine` and left `ocr_preset` alone, and `validate_convert_options` then rejects a preset that is not registered (`ocr_preset '...' is not allowed`).

That forwarding never happens on the form endpoints. `FormDepends.as_form_func` puts **every** field into `newdata`, including the ones FastAPI filled in from the field default, and passes them all to the constructor. `model_fields_set` therefore always contains `ocr_preset`, the `"ocr_preset" not in fields_set` guard is never true, and the deprecated value is dropped on the floor — accepted, unvalidated, and silently inert. The JSON body endpoints are unaffected because pydantic sees only the keys the client sent.

### The fix

Record the value FastAPI substitutes for each optional field, and skip that field in the constructor call when the submitted value is exactly that default. The model then applies its own default (same resulting value) and `model_fields_set` reflects only what the client actually sent.

After the change, on a form request:

| sent | before | after |
|---|---|---|
| `ocr_engine=tesseract` | `ocr_preset` stays `auto`, engine ignored | `ocr_preset=tesseract`, engine honoured |
| `ocr_engine=nonsense_engine` | 200, ignored | 422 `ocr_preset 'nonsense_engine' is not allowed` |
| `ocr_engine=tesseract&ocr_preset=easyocr` | `easyocr` | `easyocr` (unchanged, explicit preset still wins) |
| nothing | defaults | defaults (unchanged) |

A client that explicitly sends a value equal to the default is treated as not having sent it. That is unavoidable over multipart — the two are indistinguishable on the wire — and it is what the JSON path already does for an omitted key, so the resolved options are identical either way.

### Verification

`tests/test_form_field_defaults.py` builds a small options model with the same deprecated-field shape and drives it through `FormDepends` with `TestClient`. Against `main` 3 of the 4 tests fail (the deprecated value is not forwarded, and `fields_set` lists every field); with this change all 4 pass. The JSON-string paths for nested models and dict fields are covered too and are unchanged.

Ruff 0.9.6 (the pinned pre-commit rev) formatter and linter are clean on both touched files.

### DIFF
--- a/docling_serve/helper_functions.py
+++ b/docling_serve/helper_functions.py
@@ -4,7 +4,7 @@ import json
 import platform
 import re
 import sys
-from typing import Union, get_args, get_origin
+from typing import Any, Union, get_args, get_origin
 
 from fastapi import Depends, Form
 from pydantic import AnyUrl, BaseModel, TypeAdapter, ValidationError
@@ -65,6 +65,9 @@ def FormDepends(
     cls: type[BaseModel], prefix: str = "", excluded_fields: list[str] = []
 ):
     new_parameters = []
+    # Value FastAPI substitutes for each field when the client omits it. Used
+    # below to tell an omitted field from one the client actually sent.
+    form_defaults: dict[str, Any] = {}
 
     for field_name, model_field in cls.model_fields.items():
         if field_name in excluded_fields:
@@ -82,13 +85,20 @@ def FormDepends(
             )
         )
 
+        if not model_field.is_required():
+            form_defaults[field_name] = model_field.default
+
         # Flatten nested Pydantic models and dict/list fields by accepting them as JSON strings
         if is_pydantic_model(annotation):
             annotation = str
-            default = Form(
+            form_default = (
                 None
                 if model_field.default is None
-                else json.dumps(model_field.default.model_dump(mode="json")),
+                else json.dumps(model_field.default.model_dump(mode="json"))
+            )
+            form_defaults[field_name] = form_default
+            default = Form(
+                form_default,
                 description=description,
                 examples=None
                 if not model_field.examples
@@ -99,10 +109,12 @@ def FormDepends(
             )
         elif is_json_field(annotation):
             annotation = str
+            form_default = (
+                None if model_field.default is None else json.dumps(model_field.default)
+            )
+            form_defaults[field_name] = form_default
             default = Form(
-                None
-                if model_field.default is None
-                else json.dumps(model_field.default),
+                form_default,
                 description=description,
                 examples=None
                 if not model_field.examples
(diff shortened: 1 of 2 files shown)


dolfim-ibm (approved):
lgtm

PR #681, opened 2026-08-20
feat: expose the RQ job timeout as DOCLING_SERVE_ENG_RQ_JOB_TIMEOUT

mkBGD:
Follows up on https://github.com/docling-project/docling-jobkit/pull/227 (docling-jobkit#226), merged and released in docling-jobkit v3.4.0.

## What

Wires the new `RQOrchestratorConfig.job_timeout` through as `eng_rq_job_timeout` /
`DOCLING_SERVE_ENG_RQ_JOB_TIMEOUT`, alongside the other `DOCLING_SERVE_ENG_RQ_*` knobs.

| file | change |
|---|---|
| `docling_serve/settings.py` | `eng_rq_job_timeout: int = 3_600 * 4` next to `eng_rq_failure_ttl` |
| `docling_serve/orchestrator_factory.py` | pass it to `RQOrchestratorConfig(...)` in `_build_rq_config()` |
| `docs/configuration.md` | new row in the RQ engine table |
| `pyproject.toml` / `uv.lock` | `docling-jobkit` floor `>=3.3.0` → `>=3.4.0` |
| `tests/test_rq_job_timeout.py` | default / explicit / env-var / factory-wiring tests |

## Why

The RQ engine kills any job running longer than four hours. That limit used to be hardcoded in
docling-jobkit, so no deployment could raise or disable it — while `max_document_timeout`
defaults to seven days and `normalize_convert_options` normalizes requests that omit
`document_timeout` to exactly that maximum. A request could therefore pass policy validation
with a seven-day document timeout and still be killed at four hours by the queue backend, with
no partial result and nothing in the server or request configuration able to change it.

## Compatibility

Non-breaking: the default is `14400` seconds, exactly today's behaviour. `-1` disables the RQ
death penalty and leaves `document_timeout` as the only bound — the sensible setting for
batch/offline corpora.

## Tests

`tests/test_rq_job_timeout.py`, modelled on `tests/test_failure_ttl.py` and
`tests/test_rq_plugin_policy.py`:

- default is `3_600 * 4`
- explicit value and `DOCLING_SERVE_ENG_RQ_JOB_TIMEOUT` env var are picked up (incl. `-1`)
- `_build_rq_config()` forwards it, asserted through the existing `_CapturedConfig` patch so
  the test does not depend on the installed docling-jobkit version

Verified locally: the four new tests plus `test_failure_ttl.py`, `test_rq_queue_name.py` and
`test_rq_plugin_policy.py` → 12 passed. `ruff format`/`ruff check` clean on the touched files.

## Dependency floor

`docling-jobkit` moves from `>=3.3.0,<4` to `>=3.4.0,<4`. Older releases do not have
`job_timeout`, and pydantic ignores the unknown kwarg silently, so on those the setting would be
a no-op rather than an error. The lockfile change is limited to that one package
(3.3.0 → 3.4.0).


mkBGD:
@dolfim-ibm this is the docling-serve half of docling-jobkit#227, which you merged and released as v3.4.0 this morning. Could you approve the workflow run so CI can report? Happy to rebase if anything else is needed.

dolfim-ibm (approved):
lgtm

Issue #468, opened 2026-01-10
[BUG] can't disable OCR


crapthings:
<img width="2118" height="860" alt="Image" src="https://github.com/user-attachments/assets/a61b04c1-f11f-4964-9352-65f865c45c37" />

<img width="3984" height="1628" alt="Image" src="https://github.com/user-attachments/assets/2366fba1-c464-429d-b351-4834588edc15" />

PR #691, opened 2026-09-07
fix(ui): preserve the Enable OCR selection in conversion requests

junhyun-dev:
Unchecking Enable OCR in either Gradio conversion tab sends `options.ocr=false`,
but the API expects `do_ocr`. The unrecognized field is ignored and the default
`do_ocr=true` reaches the conversion task.

Send `do_ocr` from both callbacks. Add parameterized tests that route the real UI
callbacks to an in-process API with a mocked task queue, and verify both OFF and ON
selections at enqueue time.

Related to #468. Draft #444 replaces the UI; this change fixes the existing
Gradio paths without depending on that redesign.

Validation on Python 3.12, Serve 1.32.0, Docling slim 2.124.0:

- New regression tests: 2 failed / 2 passed before the fix; 4 passed after it.
- Related policy, form, heading hierarchy, health, and CLI tests: 78 passed
  (session asyncio loop settings for the existing health fixtures).
- Corrected file callback payload submitted to a restarted local server:
  synthetic HTML conversion succeeded and preserved the source content.
- Ruff 0.9.6 lint/format and mypy passed.

Run the regression tests with UI and development dependencies installed:
`.venv/bin/python -m pytest -q tests/test_gradio_ocr_options.py`

Browser interaction and the complete upstream CI pipeline were not run locally.


dolfim-ibm (approved):
lgtm

PR #700, opened 2026-09-17
fix(notifier): drop tasks the orchestrator no longer tracks

### LINKED ISSUE #699: WebsocketNotifier keeps entries of vanished tasks and logs "Error checking task ... status for queue position notification" on every update until restart
### Bug

`WebsocketNotifier.add_task()` creates an entry in `task_subscribers` for every enqueued task (`docling_jobkit.orchestrators.base_orchestrator.BaseOrchestrator.init_task_tracking` calls it). The entry is only removed by `BaseOrchestrator.delete_task()` -> `remove_task()`. A task can disappear from the orchestrator without that call:

1. the result is never fetched and the RQ job plus the task metadata expire (`results_ttl`),
2. the RQ zombie reaper (`RQOrchestrator._reap_zombie_tasks`) drops it from `self.tasks` without telling the notifier,
3. with more than one API replica, the result is fetched through another replica, so `remove_task()` runs on the wrong process.

From then on every pub/sub update of *any* task makes `RQOrchestrator._listen_for_updates()` call `notify_queue_positions()`, which iterates all keys, calls `task_status()` for the vanished task (one Redis GET, one RQ `Job.exists`, one INFO line `Task <id> status check`) and logs

```
ERROR docling_serve.websocket_notifier - Error checking task <id> status for queue position notification: <id>
```

The key is never dropped (`docling_serve/websocket_notifier.py:71-84` catches `Exception` and only logs), so this repeats until the process restarts. The number of lines scales with (stale entries) x (update messages).

### Numbers from a production deployment

docling-serve v1.32.0 (the notifier is unchanged in v1.33.0 and `main`), RQ engine, 2 API replicas, `results_ttl=300`, no websocket clients (all clients poll `/v1/status/poll`):

- 30 days: 2,067,289 ERROR lines for 2,953 task ids; median 571 lines and 22.5 h per task, maximum 196 h
- peak day: 495,220 ERROR lines, plus about 515,000 INFO `status check` lines that mostly came from the same notifier rounds
- example: five tasks with 2,315 lines each over 66.4 h, one line per update message (about 35/h in a quiet period)

### Expected behaviour

A task the orchestrator no longer knows should be dropped from the notifier (and any attached websockets closed) the first time `task_status()` raises `TaskNotFoundError`. No error should be logged for it again.

### Fix proposal

Catch `TaskNotFoundError` separately in `notify_queue_positions()` and `notify_task_subscribers()`, call `remove_task()` and log at DEBUG; keep the ERROR log for other exceptions. A PR with this change and five unit tests (fake orchestrator, no Redis needed; three of them fail on `main`) follows.

There is no race for fresh tasks: `RQOrchestrator.enqueue()` creates the RQ job before `init_task_tracking()` registers the task with the notifier.

### Environment

docling-serve v1.32.0 and v1.33.0, docling-jobkit 3.5.0 / 3.6.0, rq 2.12.0, upstream container image, Azure Container Apps.

### AI disclosure

The analysis, the patch and the tests were prepared with an AI coding assistant (Claude Code) under my direction. I reviewed the code paths cited above against v1.33.0 and ran the tests locally.

Ulli2060:
Resolves #699

`WebsocketNotifier` registers every enqueued task but only forgets it via `delete_task()`. When a task vanishes otherwise (result expired unfetched, zombie reaper, result fetched through another replica), the stale entry makes `notify_queue_positions()` log an ERROR on every pub/sub update until the process restarts: up to 495k lines per day in our deployment, see the issue.

Change: `TaskNotFoundError` is handled separately in `notify_queue_positions()` and `notify_task_subscribers()`. The entry is dropped, attached websockets are closed and one DEBUG line is written. Other exceptions keep the existing ERROR log. Nothing changes for tasks the orchestrator still knows.

Tests: `tests/test_websocket_notifier.py`, five cases with a fake orchestrator (no Redis needed). Three of them fail on `main`, all pass with this change. `ruff` and `mypy docling_serve` are clean.

AI disclosure: the patch and the tests were prepared with an AI coding assistant (Claude Code) under my direction. I reviewed the change against the code paths in `docling_jobkit.orchestrators.rq.orchestrator` and ran the tests locally.


dolfim-ibm (approved):
lgtm

PR #707, opened 2026-09-21
feat: add enable_api_docs and serve /metrics only with the Prometheus export

Ulli2060:
Authentication is a per-route dependency, so the schema pages (`/openapi.json`, `/openapi-3.0.json`, `/swagger`, `/docs`, `/docs/oauth2-redirect`, `/scalar`) and `/metrics` are readable without an API key on every deployment. There was no setting for the reference pages, and `DOCLING_SERVE_OTEL_ENABLE_PROMETHEUS=false` only left the Prometheus reader and the RQ/Ray collectors unregistered while the `/metrics` route stayed mounted, still publishing `python_info` and `process_*`.

- New setting `enable_api_docs` (`DOCLING_SERVE_ENABLE_API_DOCS`, default `true`). When `false`, FastAPI gets no `openapi_url`, which drops the schema, its Swagger UI and ReDoc pages and the OAuth2 redirect page, and the offline Swagger UI/ReDoc, Scalar and OpenAPI 3.0 routes are not registered either. The paths answer 404 like any unregistered route.
- `/metrics` is registered only while `otel_enable_prometheus` is `true`, which is what the configuration docs already promise for that flag.
- The CLI banner omits the docs links when they are off (and its Scalar link pointed at the ReDoc URL).

**Behaviour change:** a deployment with `DOCLING_SERVE_OTEL_ENABLE_PROMETHEUS=false` now gets 404 on `/metrics` (before: 200 with the Python process metrics only), so a scraper still pointed at it fails visibly instead of collecting an empty page. The separate `metrics_port` server is not affected and stays additive.

Design notes, in case they come up:

- 404 rather than the 403 used by `show_version_info` and `enable_management_endpoints`: an unregistered route reveals nothing about the deployment, and it is what FastAPI's own `docs_url=None` yields.
- The gate is `otel_enable_prometheus` alone, not `metrics_port`: with the flag on, the in-app route serves the same registry as the separate port. Hiding it when a port is configured would be a second behaviour change; easy to add if wanted.
- Hiding the pages instead of putting them behind the API key: the built-in OpenAPI route takes no dependencies, and a deployment that wants its schema private usually wants it invisible too.

Tests (`tests/test_api_docs_and_metrics_toggle.py`, 27 cases): all six reference routes with and without static assets (200 / 404), the API itself unaffected (`POST /v1/convert/source` -> 422), `/metrics` 200 / 404 by flag. The orchestrator dependency and the OpenTelemetry setup are stubbed as in `test_batch_endpoint.py`.

## AI disclosure

The change and its tests were prepared with an AI coding assistant (Claude Code) under my direction; I reviewed the code paths against v1.34.0 and ran the tests, ruff and mypy locally.


dolfim-ibm (approved):
lgtm

PR #711, opened 2026-09-22
feat(metrics): expose RQ queue waiting time via OTLP

nurikk:
## Summary

Add an OpenTelemetry histogram for RQ queue waiting time, measured from each attempt's enqueue timestamp to entry into worker execution, before source download or conversion.

- Emit `docling.rq.queue.wait.duration` in seconds with `rq.queue.name` as the only point attribute.
- Initialize worker OTLP metrics independently of tracing, using the existing metrics/export settings.
- Default `service.instance.id` to the RQ worker name to distinguish cumulative streams; preserve explicitly configured resource identity.
- Flush metrics on graceful worker shutdown. Retries contribute a separate observation using their refreshed enqueue timestamp.

This does not add a worker Prometheus endpoint or change the API's Redis-backed collector. The histogram describes attempts that have started, not the age of jobs still waiting.

## Configuration

On `docling-serve rq-worker`, enable `DOCLING_SERVE_OTEL_ENABLE_METRICS=true` and `DOCLING_SERVE_OTEL_ENABLE_OTLP_METRICS=true`, and configure the OTLP/gRPC receiver through the standard OTel environment variables (for example `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT`). Tracing may remain disabled.

## Validation

- Seven new integration cases passed with real Redis and a local OTLP/gRPC receiver: success, failure, retry, CLI shutdown export with tracing disabled, default/configured worker identity, and metrics/export disable flags. Set `TEST_REDIS_URL` to run these tests.
- Related regression suites: 62 passed, one pre-existing failure in `TestHealthMetricsFilterSampler.test_filtered_paths_constant` because its expected set omits `/ready`. Reproduced that failure on untouched base `a9a64c58b57f3c7673ee53dff074312b69b8b974`.
- `mypy docling_serve`, changed-file Ruff lint/format, and `git diff --check` passed.
- Reviewed using Alibaba OpenCodeReview v1.12.8 delegation mode with Claude Opus 5; no actionable defects after source reconciliation.

No production deployment or live telemetry-backend ingestion verification was performed.


dolfim-ibm (approved):
lgtm

